Issue #16, opened 2015-04-14
problam on ver 0.6


dorshay6:
when i tried to update to version 0.6 the first sms worked fine but all the other send to the same number of the first sms and with the option i provided in the end of the sms text.

pvela:
could you please turn the debugon and paste the output.

The 4th parameter to the initialize function set the debug. for ex :
nexmo.initialize("key","secret","https",true);

Also if possible paste the the code that you are using to send SMS message using the previously initialized nexmo object.

ashoksahoo:
+1
`sending message from XXXXXXXXX to +917093245428 with message Don't pay for SMS anymore. Download Mara Messenger to enjoy cool features like free outgoing SMS, doodle, chat & lots more.
sent by ashok via Mara Messenger
http://goo.gl/vKni5V
{ host: 'rest.nexmo.com',
  port: 80,
  path: '/sms/json?from=XXXXXXXXXX&to=%2B917093245428&text=Don\'t%20pay%20for%20SMS%20anymore.%20Download%20Mara%20Messenger%20to%20enjoy%20cool%20features%20like%20free%20outgoing%20SMS%2C%20doodle%2C%20chat%20%26%20lots%20more.%0Asent%20by%20ashok%20via%20Mara%20Messenger%0Ahttp%3A%2F%2Fgoo.gl%2FvKni5V&api_key=XXXXXXXXX&api_secret=XXXXXXX',
  method: 'POST',
  headers: 
   { 'Content-Type': 'application/x-www-form-urlencoded',
     accept: 'application/json',
     'Content-Length': 0 } }
POST /MaraChat/api/sendInviteSMS 200 111ms - 97b
response ended
sending message from XXXXXXXXX to +917093245428 with message Don't pay for SMS anymore. Download Mara Messenger to enjoy cool features like free outgoing SMS, doodle, chat & lots more.
sent by ashok via Mara Messenger
http://goo.gl/vKni5V
{ host: 'rest.nexmo.com',
  port: 80,
[...]

PR #17, opened 2015-04-20
Fixed Shallow Copy 

ashoksahoo:
Fixed issue #16 by changing shallow copy to deep copy.

### DIFF
--- a/lib/nexmo.js
+++ b/lib/nexmo.js
@@ -133,7 +133,7 @@ function sendMessage(data, callback) {
     } else if (!data.to) {
         sendError(callback, new Error(ERROR_MESSAGES.to));
     } else {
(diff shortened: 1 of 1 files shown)


isabo:
IMHO a more intuitive explanation is: "Modify a private copy of msgPath instead of the original, which should be treated as a constant."

i.e. the original code does not do a shallow copy -- it doesn't copy the object at all!

pvela:
Thanks for pointing out the problem and the pull request.

ashoksahoo:
It will get updated to npm automatically or it takes a while?

pvela:
Its updated in NPM now. This issue existed in other functions as well, fixed all of them. Thanks again.

PR #26, opened 2015-11-25
We now always use HTTPS to communicate with Nexmo

damienwhaley:
I've changed the library so that it will now always communicate over HTTPS to Nexmo.

Nexmo have made an announcement here about it:

https://help.nexmo.com/hc/en-us/articles/213877768

As part of this I've also fixed up the test suite so it can be run, and I've updated the README to include testing instructions, and the change in the initialisation method.

Note that this is a breaking change as the API signature for the init() method has changed.


pvela:
Thanks for PR @damienwhaley, the new version is published to NPM.

PR #34, opened 2016-03-15
Enhanced code for testing purposes

vsadokhin:
Hello

I've made these changes in code that should not change the library behavior if new setHost or setPort are not used. Changes goal is make real library calls to local machine or CI server instead of real nexmo API while acceptance and integration tests are run. List of changes: 
- Added port variable, default value is 443
- Changed code to use port variable
- Added require http module
- Added setHost and setPort functions to override default values externally
- Changed sendRequest function to use http if port is not 443

Let me know if there is a problem with the changes and how you like me improve it. 

Thank you,
Vasiliy


pvela:
Can you please add a section in readme describing why you would use this and how to use it.

vsadokhin:
I enhanced Testing section at the end of the readme. Please, have a look and let me know if any.

PR #36, opened 2016-04-06
README badge and title updates

leggetter:
- Removed build status since there are no tests that can be run as part of a build
- Added NPM details via badge (dependencies, version etc.)


leggetter:
Thanks @pvela :+1:

PR #43, opened 2016-05-04
For loop, and cancelNumber bug fixes

### LINKED ISSUE #42: Runtime error using searchNumbers
I'm getting an error on this line: https://github.com/Nexmo/nexmo-node/blob/9136d8471ea27487bdec95f1bdd78261a1afe621/lib/nexmo.js#L290

```
.../easynexmo/lib/nexmo.js:290
            for (arg in pattern){
                 ^

ReferenceError: arg is not defined
```

Seems arg should be declared earlier or inline to be valid JS.

cbetta:
Closes #42 and #44

### DIFF
--- a/lib/nexmo.js
+++ b/lib/nexmo.js
@@ -287,7 +287,7 @@ exports.searchNumbers = function(countryCode, pattern, callback) {
             callback = pattern;
         } else if (typeof pattern == 'object'){
             searchEndpoint.path = searchEndpoint.path + '&';
-            for (arg in pattern){
+            for (var arg in pattern){
                 searchEndpoint.path = searchEndpoint.path + arg + '=' + pattern[arg] + '&'
             }
         } else {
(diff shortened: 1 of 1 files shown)


leggetter:
Does this fix the previous exception? I would have thought `arg` would just
be a global.

cbetta:
Yes it does. Added reference to that post.

cbetta:
Added another fix for issue #44.

lib/nexmo.js
```diff
@@ -287,7 +287,7 @@ exports.searchNumbers = function(countryCode, pattern, callback) {
             callback = pattern;
         } else if (typeof pattern == 'object'){
             searchEndpoint.path = searchEndpoint.path + '&';
-            for (arg in pattern){
+            for (var arg in pattern){
```
leggetter, line 290: I still don't get how this fixes anything. It's still a valid improvement, but I'm befuddled about the fix :)

See: http://jsbin.com/mihohu/edit?js,console

The `var` isn't required in the `for` loop.

cbetta, line 290: It is in my runtime (node, commandline, piped through babel). I think it's technically _not_ correct JS.

leggetter, line 290: No - it's allowed. Just hoists the variable to a global.

It'll probably be the Babel transpilation.

We should probably have have JSLint/JSHint/ESLint. For later!

cbetta, line 290: So happy to merge?

pvela, line 290: Couldn't jumpin earlier, I am good with this merge.

PR #48, opened 2016-05-09
Add App Beta API CRUD

cbetta:
Left out README for now as the API is still on flux and only used internally.
- Add getApplications()
- Add getApplication()
- Add createApplication()
- Add updateApplication()
- Add deleteApplication()
- Add updateApplication()
- Prevent JSON parsing when the response is empty for a DELETE call


pvela:
Looks good. Should we hold off on updating readme until the API's are out of beta?

cbetta:
Agreed. @leggetter ^

PR #55, opened 2016-05-16
Namespaced ES2015 classes mapping to existing `nexmo` object (fixed rebase)

leggetter:
Replaces #54 

This introduces a simple set of ES2015 classes that use the existing `nexmo` object for functionality.

Main changes are:
- Package name updated to `nexmo`
- Version is 1.0.0-beta-x (where `x` is being updated as bugs are found)
- To use you now create an instance of a `Nexmo` object and access functionality off of namespaced properties:
  - `nexmo.message`
  - `nexmo.voice`
  - `nexmo.number`
  - `nexmo.numberInsight`
  - `nexmo.account`
  - `nexmo.verify`
  - `nexmo.app`
- Some API names have been simplified e.g. `nexmo.numberInsightBasic` -> `nexmo.numberInsight.get({level:'basic'})`
- The package uses ES2015 and Babel for transpilation

Tests have been written to make sure that all functionality is still exposed and that the APIs of v0.x are mapped to v1.0. See the test files for details of the mapping.

Also see the test files for details of how the functionality has been grouped into namespaces. It would be great to get some feedback on this last point.


cbetta:
If this is a major version bump,  why still support the older easy Nexmo
API? Sounds like a good moment to deprecate that API.

leggetter:
> If this is a major version bump,  why still support the older easy Nexmo
> API? Sounds like a good moment to deprecate that API.

If folks want to use the old API they would need to:

``` js
var nexmo = require('nexmo/lib/index');
```

to get hold of the older reference.

So we have kind of deprecated the older API since we're saying do:

``` js
var Nexmo = require('nexmo');

var nexmo = new Nexmo({key: KEY, secret: SECRET});
// use namespaced objects to interact with the Nexmo platform
```

Maybe I'm missing something. What would you suggest, @cbetta ?

cbetta:
I say if people want the old API they do require easynexmo. Why keep the
option of the old stuff? Simplify the code base while you can.

leggetter:
> I say if people want the old API they do require easynexmo. Why keep the
> option of the old stuff? Simplify the code base while you can.

The point of approaching things this way (still using the old `easynexmo`) functionality was to avoid the big bang approach; refactoring a tonne of code in one go.

The idea with this release is to get people using the new namespaced library/API and `nexmo` package ASAP. Then we can look at migrating the code within `lib/index.js` into the appropriate classes.

There won't be anything telling people to do `require('nexmo/lib/index/js')` (with the exception of this PR thread). As far as anybody using the `nexmo` package is concerned there is only the new namespaced API.

cbetta:
Agreed. Sounds good. My main concerns would be to simplify the codebase and that involves not promising for old APIs to remain working. Would be good to add a deprecation warning for the OLD API to warn people this won't remain in the version after next.

README.md
```diff
@@ -1,226 +1,301 @@
...
-Usage :
-=======
+## Usage
 
 ```js
-var nexmo = require('easynexmo');
+var Nexmo = require('nexmo');
 
-nexmo.initialize(KEY, SECRET, DEBUG);
+var nexmo = new Nexmo({key: KEY, secret: SECRET}, {debug: DEBUG});
```
cbetta: Hmm, my guts say name these after they same variables used in the API: `api_key` and `api_secret`.

examples/pre-v1.js
```diff
@@ -1,5 +1,7 @@
 'use strict';
 
+require('dotenv').config({path:__dirname + '/.env'});
```
cbetta, line 3: Why a sample that uses dotenv? Far from MVP is it?

leggetter, line 3: Purely because it's a handy way of running samples.

leggetter:
> Would be good to add a deprecation warning for the OLD API to warn people this won't remain in the version after next.

It's JS so there isn't really a good way of doing this.

We could JSDoc the codebase with `@deprecated`? But that won't trigger any warnings if people try to use them.

cbetta:
`console.warn`?

leggetter:
> console.warn?

We can't use `console.warn` as the old API is still being used by the new namespaced code. It's just the old API should be classed private.

I kinda think this is a non-issue. If we don't tell people about `nexmo/lib/index` then the API won't be used. I'd be happy to mark them as `@private` for anybody delving into the code.

README.md
leggetter: I prefer the brevity of `key` and `secret`.

I'm also not particularly keen on underscores in JS variables. I think the general JS concensus is camelCase https://github.com/airbnb/javascript#naming-conventions (I wasn't going to mention that about the `nexmo-cli` codebase 😼 ).

cbetta: The `nexmo-cli` _should_ be camelCase where possible, except for where it maps the command line input to the commander library ;) I might have broken this rule in places though, maybe should add a lint rule for it :)

As for `api_key` vs `key` I guess my thoughts are as followed. Until now the library was a thin wrapper around the API. Params had the same name in the lib as in the API (though not consistently). Personally I always prefer my lightweight wrapper lib to match the API in namign as far as possible as it means I can make assumptions based on the API, not on the lib.

But hey this PR seems to break a lot of that 1:1 mapping anyway, so if that's the goal then changing it to the shorter names is fine with me :)

examples/pre-v1.js
cbetta, line 3: Handy, but not required. I'm a strong believer in less is better :) See https://betta.io/blog/2016/04/10/minimum-viable-samples/

Anyway it's your code so your decision. If 90%+ of Node devs are familiar with `dotenv` then leave it in, otherwise I'd scrap it.

cbetta:
Probably not worth trying to detect how a lib is included. Leave it.

examples/pre-v1.js
leggetter, line 3: `.env` isn't required. The key/secret can be edited directly into the sample code. IMO it's pretty clear/simple and `.env` is becoming very common practice across runtimes.

But it is an interesting point and I'll look out for similar comments and problems that it creates.

leggetter:
@pvela when you get a moment could you please cast your eyes over this PR?

PR #66, opened 2016-09-29
Replace NI Advanced call with synchronous version

### LINKED ISSUE #64: Add Insights Advanced sync callback
Necessary for nexmo/nexmo-cli#87

cbetta:
- Moved async call into own option
- Fixed all API endpoints to ones used in current docs
- Updated documentation to highlight changes

This will probably require a minor release bump as this commit will break for anyone expecting `nexmo.numberInsight.get({level: 'advanced'})` to be asynchronous.

Closes #64


src/index.js
```diff
@@ -13,8 +13,9 @@ var checkVerifyEndpoint = {host:'api.nexmo.com',path:'/verify/check/json'};
 var controlVerifyEndpoint = {host:'api.nexmo.com',path:'/verify/control/json'};
 var searchVerifyEndpoint = {host:'api.nexmo.com',path:'/verify/search/json'};
 var niEndpoint = {host:'rest.nexmo.com',path:'/ni/json'};
-var niBasicEndpoint = {host:'api.nexmo.com',path:'/number/format/json'};
-var niStandardEndpoint = {host:'api.nexmo.com',path:'/number/lookup/json'};
+var niBasicEndpoint = {host:'api.nexmo.com',path:'/ni/advanced/async/json'};
```
leggetter: The variable name and `path` value don't seem to match up.

src/NumberInsight.js
```diff
@@ -33,55 +33,59 @@ class NumberInsight {
...
     delete options.level;
-    
+
     if(level === 'advanced') {
+      this._nexmo.numberInsightAdvanced.apply(this._nexmo, arguments);
+    } else if (level === 'asyncAdvanced') {
       this._nexmo.numberInsight.apply(this._nexmo, arguments);
     }
     else if(level === 'standard') {
+      console.log('3');
```
leggetter: Debug left in.

src/NumberInsight.js
```diff
@@ -33,55 +33,59 @@ class NumberInsight {
...
       this._nexmo.numberInsightStandard.apply(this._nexmo, arguments);
     }
     else {
+      console.log('4');
```
leggetter: Debug left in.

README.md
```diff
@@ -347,6 +349,20 @@ nexmo.number.buy(countryCode, msisdn, callback);
 ```
 
 ### Cancel Number
+=======
```
leggetter: Is this required? The `###` should be enough and I think all other headers use this syntax.

README.md
```diff
@@ -333,6 +333,8 @@ nexmo.number.search(countryCode,options,callback);
 	* `size`
 
 For more details on what the above options mean refer to the Nexmo API  [documentation](https://docs.nexmo.com/tools/developer-api/number-search)
+=======
```
leggetter: Required?

README.md
leggetter: I'm not sure why this seems to be followed by code. Maybe the GitHub preview has gone wonky?

I see:

``````
### Cancel Number
======

nexmo.numberInsight.get({level: 'advanced', number: NUMBER}, consolelog);
\```

For more info...
``````

(ignore the backslash)

**Review by leggetter — CHANGES_REQUESTED** [review]
There seem to be quite a few whitespaces changes in the code. Not sure why that is, and I'm not going to get to worried about it :)

Main thing is that the README seems to have been broken. I think a bunch of text may have accidentally been removed.

File: README.md (hunk 5 above)
Comment (leggetter) [reply README.md]: I think this is where a mistake has been made and text accidentally removed.

**Comment by cbetta** [comment]
@leggetter Damn, that's not right. Seems my rebase didn't go well. I'll have a look

File: README.md [hunk 6]
```diff
@@ -210,23 +210,33 @@ nexmo.numberInsight.get({level: 'basic', number: '1-234-567-8900'}, consolelog);
...
 ```js
-nexmo.numberInsight.get({level: 'standard', number: '1-234-567-8900'}, consolelog);
+nexmo.numberInsight.get({level: 'standard', number: '1-234-567-8900'}, callback_function);
 ```
 
 ### Advanced
 
 ```js
-nexmo.numberInsight.get({level: 'advanced', number: NUMBER}, callback);
+nexmo.numberInsight.get({level: 'advanced', number: NUMBER}, callback_function);
```
leggetter: `callback_function` -> `callback` - Looks like your rubyism has snuck in!

README.md
```diff
@@ -210,23 +210,33 @@ nexmo.numberInsight.get({level: 'basic', number: '1-234-567-8900'}, consolelog);
...
 ```
 
 For more information check the documentation at https://docs.nexmo.com/number-insight/advanced
 
+### Advanced Async
+
+Number Insight Advanced might take a few seconds to return a result, therefore the option exist to process the result asynchronously through a webhook.
+
+```js
+nexmo.numberInsight.get({level: 'advancedAsync', number: NUMBER, callback: "http://example.com"}, callback_function);
```
leggetter: just `callback` is cool

leggetter (commented):
I'd suggest we stick with the consistent snakeCase naming convention of variables. I can live with the whitespace changes :)

README.md
cbetta: @leggetter haha, sorry. Was just trying to clarify here there's a difference in the Async call between the callback URL for the webhook and the callback function for this function.

Want me to change it to `callbackFunction`?

Additionally we should add this to the linter to make sure we ensure camelCase.

README.md
cbetta: See other comment

README.md
leggetter: nah - just use `callback`. All/most of the other examples just use that.

README.md
cbetta: Renamed it to `callback` everywhere

README.md
cbetta: @leggetter fixed. And updated all other mentions to callback too instead of `consolelog` which seems to not be used anywhere else.

leggetter (approved):

leggetter:
@cbetta - so, now a version bump and the need to publish?

@tjlytle - worth your input on the decision we make for the version. There's a breaking change here with one of the APIs since insight advanced used to be async. It's now sync. There's a specific async version now.

tjlytle:
@leggetter @cbetta Any way to keep the existing behaviour in _this_ version and change in a future major release? `advancedSync` would be the new `level`.

cbetta:
@leggetter @tjlytle I'd say bump it to `1.1.0`. This is what semver is for.

leggetter:
If it's a breaking API change then in theory we should be changing the MAJOR. So that's 2.0.0.

cbetta:
Ah I stand corrected. Yeah so technically 2.0.0. I'm happy with that. Embrace Chrome's "don't be afraid to increase the major" motto!

Also happy to make the change minor ;)

tjlytle:
I'd rather save 2.0 for a more significant change of the library (which
will likely be sooner than later).

cbetta:
@tjlytle although I understand your intent I also want to point out that that's a completely cosmetic decision. Version numbers are meant to signify change in a semantic way, not to celebrate some milestone in the code base. 

Anyway, don't want to be a blocker so I'm happy to change the API to `advanced` and `advancedSync` to keep the code backwards compatible. @leggetter thoughts?

PR #69, opened 2016-09-30
Add a way to generate a JWT with a custom key, app ID and additional claims

### LINKED ISSUE #68: Add way to generate JWT with App ID and Key to be used from outside module
This is needed for Nexmo/nexmo-cli#88 

@leggetter you already added a `JwtGenerator` but I can't use this from the CLI as it's not being exported yet by the module. What do you reckon we do, export the class or an instance?

cbetta:
Closes #68 

Have settled for just a simple change as defined by @leggetter in #68.


leggetter (commented):
Worth providing a README update to show how to do this as part of this PR?

cbetta:
@leggetter not sure re README. The existing JWT behaviour wasn't documented either and personally I don't think I quite understand what the JWT would be used for to write a good doc.

leggetter:
Ok, I'll add something.

It raises a good point - that there's no clear way to directly use a JWT right now. Maybe the libraries should be updated to support using JWT upon initialisation in the same way that they should support `applicationId` and `privateKey`.

It needs some thought.

There are other uses for the JWT outside of the client libraries. We've just not exposed them yet.

Leave this with me.

leggetter:
@cbetta based on this update https://github.com/Nexmo/nexmo-cli/issues/88 could use:

``` js
import Nexmo from 'nexmo';

var token = Nexmo.generateJwt(privateKey, listOfOtherClaimsMap);
```

.editorconfig
```diff
@@ -0,0 +1,7 @@
+root = true
+
+[*.js]
+trim_trailing_whitespace = false
```
cbetta: Don't we want this to be true?

.eslintrc.js
```diff
@@ -29,10 +29,10 @@ module.exports = {
             'warn'
         ],
         'no-trailing-spaces': [
-          'error'
+          'error', { "skipBlankLines": true }
```
cbetta: Why? I prefer blank lines to actually be empty, not a long list of spaces....

leggetter: If you could push a editorconfig that "makes this happen" (Ideally in Atom) then I'd be happy to use it. To be honest, the eslint stuff has actually been getting in the way of getting stuff done so I'm just looking for a fast and efficient solution that helps us do stuff in the right way.

cbetta: @leggetter check the (default) `whitespace` package in atom. I think those options by default should be on. Alternative if we add `trim_trailing_whitespace = true` to the `.editorconfig` it should just work.

I know linting can feel like a pain, but in the end it helps limit technical debt and it helps dead code, potential errors, and much more. I do agree that adding a good `.editorconfig` would probably improve things a lot.

.editorconfig
```diff
@@ -0,0 +1,7 @@
+root = true
+
+[*.js]
+trim_trailing_whitespace = true
+insert_final_newline = false
```
cbetta: This is generally better to set to true. Even in JS (plenty of articles out there making a case for this)

PR #71, opened 2016-09-30
Make advanced and advancedSync backwards compatible

cbetta:
Following our discussion in #66 I've renamed the levels and added a deprecation warning for users so we can change the `advanced` level in 2.0.

@tjlytle you agree this works?

### DIFF
--- a/src/NumberInsight.js
+++ b/src/NumberInsight.js
@@ -71,10 +71,11 @@ class NumberInsight {
     // remove 'level' as it's a library-only parameter
     delete options.level;
 
-    if (level === 'advancedAsync') {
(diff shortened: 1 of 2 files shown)


tjlytle:
@cbetta Looks great, thanks!

cbetta:
@leggetter mind pushing this change to NPM?

leggetter:
@cbetta I'd like to get the JWT stuff #69 (scope has crept, but it needs to) in before pushing a new release.

cbetta:
@leggetter sure thing, though I personally think it's artificial to wait. 

As for #69 I'll wait for you and Tim to come down to a bit more of a spec on that one.

PR #73, opened 2016-10-03
Linted source

cbetta:
Work in progress to lint the source as well (tests are already linted and merged). Mostly fixable except for 2 issue. See below

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -3,22 +3,65 @@
 var querystring = require('querystring');
 
 var initialized = false;
-var msgpath = {host:'rest.nexmo.com',path:'/sms/json'};
-var shortcodePath = {host:'rest.nexmo.com',path:'/sc/us/${type}/json'};
-var ttsEndpoint = {host:'api.nexmo.com',path:'/tts/json'};
-var ttsPromptEndpoint = {host:'api.nexmo.com',path:'/tts-prompt/json'};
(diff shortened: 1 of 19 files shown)


src/index.js
```diff
@@ -616,11 +665,11 @@ exports.setHost = function(aHost) {
...
-}
+    niBasicEndpoint.host = aHost;
+    niStandardEndpoint.host = aHost;
+    applicationsEndpoint.host = aHost;
+};
 
 exports.setPort = function(aPort) {
-  port = aPort;
-}
+    port = aPort;
```
cbetta, line 674: ESLint says that `'port' is not defined` which is true. It also seems `port` is not used anywhere either. Should we remove this or properly implement this?

src/index.js
```diff
@@ -3,22 +3,65 @@
 var querystring = require('querystring');
 
 var initialized = false;
```
cbetta, line 5: ESLint claims `'initialized' is defined but never used` even though it is on line 109. We should see and test if this actually works. It seems to me like JS is assigning a local variable on line 109, not a global one. But I'm not sure if this is an issue with ESLint or not.

src/index.js
cbetta: @leggetter i think this is probably an example of some dead code linting has found.

leggetter:
Let's kill this one for now.

PR #80, opened 2016-10-25
Accept private key as string

### LINKED ISSUE #79: Support private key as authentication vs file?
I currently find the library to be a bit inconsistent when it comes to private keys. As a user I am left to write a key to file:

``` js
var Nexmo = require('nexmo');

var nexmo = new Nexmo({
    apiKey: 'key',
    apiSecret: 'secret',
});


nexmo.app.create(
  'First Voice App',
  'voice',
  'http://example.com/answer',
  'http://example.com/event',
  {},
  function(err, res) {
    var appId = res.id;
    var privateKey = res.keys.private_key;
    var privateKeyFile = __dirname + '/' + appId;
    require('fs').writeFileSync(privateKeyFile, privateKey);
  }
);
```

But then when I try to configure the library with a key it requires a filename and reads the file:

``` js
  nexmo = new Nexmo({
      apiKey: 'key',
      apiSecret: 'secret',
      applicationId: appId,
      privateKey: privateKeyFile
    });
```

I somehow think it would make a lot more sense for the initializer to just accept the key itself and not bother itself with reading from disc. This would also make it possible to create an app and then reinitialize the library in a few lines of code without having to write the key to disc.

Final note: I might want to save my API key somewhere else on the disk (another volume, database, etc) so having this option would be useful regardless.

Thoughts @leggetter @tjlytle ?

cbetta:
Closes #79

### DIFF
--- a/src/Credentials.js
+++ b/src/Credentials.js
@@ -11,8 +11,9 @@ import JwtGenerator from './JwtGenerator';
  * @param {string} apiKey - A Nexmo API Key
  * @param {string} apiSecret - A Nexmo API Secret
  * @param {string|Buffer} [privateKey] -  When a string value is passed it should
- *                        represent the path to the private key. If a Buffer is
- *                        passed then it should be the key read from the file system.
+ *                        either represent the path to the private key, or the actual
+ *                        private key in string format. If a Buffer is passed then
+ *                        it should be the key read from the file system.
  */
 class Credentials {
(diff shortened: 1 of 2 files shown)


leggetter:
Merged in https://github.com/Nexmo/nexmo-node/commit/ea764cce7407809a2936080394c626e69aaafe2f

PR #86, opened 2016-12-18
Add examples for talk resource

sprusr:
There were no examples for using the `/v1/calls/{uuid}/talk` endpoint, something which I would have found very useful for getting a project of mine up and running. Done in the same format as similar existing examples.


leggetter:
Thanks @sprusr 👍

PR #99, opened 2017-01-24
Fix lint warnings

### LINKED ISSUE #67: Add active linting to tests
We should add linting to our tests and ensure all linting passes before code is merged. This will initially require one big refactor to lint all the current code. Things we should look at fixing:
- Strip trailing white space
- Pick tabs or spaces - probably 4 spaces
- Pick whether to enforce trailing semi colons
- Enforce camel case over snake case

cbetta:
Test linting was already in place but failing due to most of the code using 4 spaces and the linter expecting 2. By changing the linter to 4 and fixing 4 lines of code we now got rid of all warnings.

Closes #67

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -11,7 +11,7 @@ module.exports = {
     'rules': {
         'indent': [
             'warn',
-            2
+            4
         ],
         'linebreak-style': [
             'error',
--- a/test/Nexmo-test.js
+++ b/test/Nexmo-test.js
@@ -217,11 +217,11 @@ describe('Nexmo Object instance', function() {
 
         var expectedJwt = Nexmo.generateJwt(
           privateKey,
-          {
-              'application_id': appId,
-            'iat': iat,
-              'jti': jti
-          }
+            {
+                'application_id': appId,
+                'iat': iat,
+                'jti': jti
+            }
         );
 
         var nexmo = new Nexmo({


.eslintrc.js
```diff
@@ -11,7 +11,7 @@ module.exports = {
     'rules': {
         'indent': [
             'warn',
-            2
+            4
```
leggetter, line 14: Since we want the project to have an indent of `2` and not `4` I think it's best to keep this at 2 and clearly indicate to others (and their editors - should we have a `.editorconfig`) that the project uses 2. Changing this to 4 really just hides that we haven't fixed the indentation yet.

leggetter (commented):
IMO we close this and fix the indentation across the project instead.

.eslintrc.js
cbetta: Sure, but for now this will stop lint warnings and at least keep the project consistent within itself. We can then do a migration to 2 spaces next.

cbetta:
@leggetter coming back to this. I really think we should merge this for now because as is the project  is a pain to work with. Currently running tests is overloaded by the warnings, and as a result it's hard to actually see any other linting error or warning, allowing for new inconsistencies to creep in. Either we remove the rule or fix the code one way or another.

leggetter:
Should be fixed by https://github.com/Nexmo/nexmo-node/pull/107

PR #104, opened 2017-01-30
Bump to 1.1.2 - fixes JSON parse error

cbetta:


src/HttpClient.js
```diff
@@ -80,11 +88,15 @@ class HttpClient {
...
+                      return callback({
+                        message: "The API response could not be parsed.",
+                        parseError: parsererr
+                      }, null);
 	                }
                 }
 
                 if(response.statusCode < 200 || response.statusCode > 299) {
-                  err = retJson;
+                  callback(err, retJson);
```
leggetter: Could this result in two callbacks being made? After this `callback` is executed the code would run to the next `callback`.

leggetter (dismissed):
In general the code is now a bit harder to follow IMO since it has 3 (or maybe even 4 if `return` is added before all `callback` invocations) points of return.

Can it be simplified?

src/HttpClient.js
cbetta: Yes it would. I will fix.

cbetta:
@leggetter yes it could be, by adding actual tests to this class. Currently there are none.

cbetta:
Just kidding. Rewriting the whole thing and adding a few tests.

cbetta:
@leggetter mind giving he new version some fresh eyes?

leggetter (approved):

PR #105, opened 2017-01-31
Add method for getting and storing files

### LINKED ISSUE #85: Add support for retrieving call recordings
For @jonlingard :)

Generally missing across the board on docs and in other libraries.

cbetta:
Closes #85 

Adds 2 methods:

```js
nexmo.files.get(fileIdOrUrl, function(err, resp) {
  // buffer or similar
});
```

```js
nexmo.files.save(fileIdOrUrl, fileName, function(err, resp) {
  // callback with fileName
});
```

## Additionally

* Updated `HTTPClient` to use Buffers until it knows the content is JSON, alternatively returns a buffer.


cbetta:
@leggetter thoughts on this?

src/FilesResource.js
```diff
@@ -0,0 +1,81 @@
...
+  }
+
+  /**
+   * Save remote File locally
+   *
+   * @param {string} [fileIdOrUrl] - The unique identifier or URL for the file
+   * @param {string} [fileName] - Path name to store the file to
+   * @param {function} callback - function to be called when the request completes.
+   */
+  save(fileIdOrUrl, fileName, callback) {
```
leggetter: `fileName` or `filePath`?

src/HttpClient.js
```diff
@@ -59,16 +59,27 @@ class HttpClient {
 
     request.end(endpoint.body);
 
-    var responseData = '';
+    var responseData = [];
```
leggetter: Is this an Array of Buffers or an Array of strings? Worth a comment?

src/HttpClient.js
```diff
@@ -85,7 +96,7 @@ class HttpClient {
 
   }
 
-  __parseReponse(status, data, method, callback) {
+  __parseReponse(status, contentType, data, method, callback) {
```
leggetter, line 102: Thought:

Passing the content type rather than checking the `typeof data`? `Buffer` vs `string`

Maybe it's more explicit the way it's been implemented. It's just the elements in `responseData` weren't too clear earlier.

leggetter (commented):
A few comments I'd like your thoughts on. Generally LGTM 👍

src/FilesResource.js
cbetta: Whatever you want. You prefer `filePath`?

src/HttpClient.js
cbetta: I could have passed in `isBinary` but this is a lot more explicit and flexible if more content types would arise.

src/HttpClient.js
cbetta: Sure could add a comment

src/FilesResource.js
leggetter: I think `filePath` is clearer, yes.

leggetter: Actually - should it be a filename or file descriptor? Following the convention set by `fs.writeFile`

cbetta:
@leggetter have added the changes you requested

src/FilesResource.js
leggetter: So, maybe it should be `file` which can be of either type. see https://nodejs.org/api/fs.html#fs_fs_writefile_file_data_options_callback

cbetta:
@leggetter also: do we want this added to the CLI next?

leggetter (approved):
LGTM :shipit: 

cc: @girliemac

PR #110, opened 2017-02-26
Check HTTP response `statusCode` and not `status`

poying:
HttpClient does not pass error object to callback, because `status` variable always be `undefined`.


ghost:
Hum

leggetter:
@poying thanks for flagging this. I've merged this in to #115

PR #119, opened 2017-03-21
Add RTC support to application create/update

### LINKED ISSUE #118: Add support for RTC type application
Context: https://github.com/Nexmo/nexmo-cli/issues/108

It seems there is a new application type called `rtc` that is not yet documented by requested to provide support for regardless. Currently the code validated the param for the app type and rejects `rtc`.

cbetta:
Closes #118 

Add the `rtc` application type to the application creation/updating API calls.

Related to: https://github.com/Nexmo/nexmo-cli/issues/108


leggetter (commented):
Other than validating clearly wrong values (e.g. empty strings) maybe we should let the API check potentially valid values.

PR #121, opened 2017-03-22
Remove number and application type validation

### LINKED ISSUE #118: Add support for RTC type application

### LINKED ISSUE #120: Add support for numbers shorter than 10 digits
As @sammachin pointed out the current library incorrectly assumes that all numbers are > 10 digits. This is a) incorrect, and b) probably not a check the library should do. 

For now I will update the library to just check that a numbers is a) just digits and b) 1 digit or more.

cbetta:
Closes #120 
Closes #118
Replaces #119 

Changes all number validation to not limit numbers to 10 digits or more. Not only is this incorrect, but also doesn't actually validate if the number is in any way digits.

A regex for the validation already existed so this has now been updated.

I do not foresee any issues with this code, as realistically it should not exist in the SDK regardless, as the API will catch these errors already too. 

This code has no tests as the `index.js` has no tests as it stands.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -19,6 +19,7 @@ var niAdvancedEndpoint = {host:'api.nexmo.com',path:'/ni/advanced/json'};
 var applicationsEndpoint = {host:'api.nexmo.com',path:'/v1/applications'};
 var up = {};
 var numberPattern = new RegExp("^[0-9 +()-]*$");
+
 var _options = null;
 
 //Error message resources are maintained globally in one place for easy management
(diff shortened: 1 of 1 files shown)


leggetter (commented):
Is there a benefit of validating in the client over letting the API doing the validation?

IMO we should consider passing that on to the API and removing validation from the client.

cbetta:
@leggetter agreed. Just didn't want to break too much in 1 go, but tbh that makes the most sense. Ok if I deploy that change and release 2.1.0 (or 2.0.2)? We're in a slight hurry for @sammachin to test something.

cbetta:
Changed code to remove the check where possible. Also removed check for Application type. 

There is one bit of code still where the regex is used, around the `insightCommon` method. I have left this in place as it's not directly related to what we're doing here.

leggetter (approved):

PR #125, opened 2017-04-26
Prevent HttpClient modify default header settings

poying:


mheap:
@poying I've just merged this in to master - thanks for your contribution!

I added a test for this behaviour too in #162

PR #126, opened 2017-05-12
Add ability to generate a sub token with no restrictions

TheChrisVela:
Add a jwt example
Modify jwt test

With this change, users will be able to generate a sub token with no restrictions.

Thank you @jacopomaroli for your help and suggestions for this change.

In addition to modifying the tests, I've also verified through https://jwt.io/ that a jwt generated with this will include in its payload:

```json
"acl": {
    "paths": {
      "/**": {}
    }
}
```

### DIFF
--- a/examples/ex-generate-jwt.js
+++ b/examples/ex-generate-jwt.js
@@ -1,0 +1,16 @@
+module.exports = function(callback, config) {
+
+  var Nexmo = require('../lib/Nexmo');
+
+  var nexmo = new Nexmo({
+      apiKey: config.API_KEY,
+      apiSecret: config.API_SECRET,
+      applicationId: config.APP_ID,
+      privateKey: config.PRIVATE_KEY,
+    },
+    {debug: config.DEBUG}
+  );
+
+  var jwt = nexmo.generateJwt();
+  console.log(jwt)
+};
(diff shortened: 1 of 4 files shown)


TheChrisVela:
If it's better that generating a token with no restrictions is optional that you have to enable with a  flag let me know.

leggetter (commented):
If acl is required we are breaking backwards compatibility 🚨

Or is it required for the conversation API, but it not being present will not matter to the Voice API?

TheChrisVela:
It's required for conversation API. Not sure how this will affect Voice API. Maybe @sammachin knows?

leggetter:
@ChrisGuzman but shouldn't we just insist that it's provided as part of the JWT generation rather than setting a default? Would providing a default that gives access to "all paths" be a security concern?

sammachin:
Yes, tokens should be scoped to provide the narrowest path possible and
with the shortest practical TTL, otherwise you are just creating additional
account passwords.

I havnt’ looked at the specifics but in general for VAPI you don’t actually
need to see the token its just generated and used as part of the API call
so the library should be able to set the appropriate path at that point as
it will know what the token is to be used for before its generated.

TheChrisVela:
Chatted with @sammachin and we agreed this isn't the best way to implement this. Closing the PR

PR #136, opened 2017-07-09
Add note about providing private key inline

### LINKED ISSUE #131: Add private key support
Currently the nexmo-node library takes the private key as a path, this presents an issue in certain scenarios where the private key is to not be checked into version control; for example open source projects.

If for instance using hosting such as Heroku your application would need to:

1. Read the private key from the environment.
2. Write the private key to disk
3. Initialize `Nexmo` with the the private key path.

If the library were to support providing the private key itself then step 1 & 2 could could be avoided.

benfoxall:
It looks like a [buffer can be used](https://github.com/Nexmo/nexmo-node/blob/master/src/Nexmo.js#L106-L114) to set the private key when constructing the wrapper.

This PR adds a note of that in the readme.

Resolves #131

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,7 +30,7 @@ var nexmo = new Nexmo({
 * `apiKey` - API Key from Nexmo
 * `apiSecret` - API SECRET from Nexmo
 * `applicationId` - The Nexmo Application ID to be used when creating JWTs. Required for voice related functionality.
-* `privateKey` - The path to the Private Key to be used when creating JWTs.  Required for voice related functionality.
+* `privateKey` - The Private Key to be used when creating JWTs; either as a path to the key file, or a [Buffer](https://nodejs.org/api/buffer.html#buffer_class_method_buffer_from_string_encoding) containing the file contents.  Required for voice related functionality.
 * `options` - Additional options for the constructor
 
 Options are:


AverageMarcus:
Might also be worth adding that a string can be used too containing the text contents of the key. Useful if using env vars to store the key contents.

benfoxall:
👋 @AverageMarcus!

Looking at [the source](https://github.com/Nexmo/nexmo-node/blob/master/src/Nexmo.js#L106-L108) - I think if you provide the a string with the text contents it would throw a `File <<FILE STRING>> not found.` error.  Do you not get that error?

Wrapping the string in `Buffer.from` will let you provide the file contents though:

```js
var nexmo = new Nexmo({
  /* ... */
  privateKey: Buffer.from(process.env.PRIVATE_KEY_CONTENTS)
})
```

(That's actually what I was trying to describe in this PR.  Maybe it could be clearer?)

AverageMarcus:
Take a look here where the creds are parsed: https://github.com/Nexmo/nexmo-node/blob/master/src/Credentials.js#L30

benfoxall:
Ah, nice one. I wonder why that wasn't working for me when I raised this before 🤷‍♂️

mheap:
Thanks for the contribution - I'll add a note about using a string to the README too

PR #146, opened 2017-09-25
wip e2e test

AlexLakatos:
run it with `npm run compile && mocha --compilers
./node_modules/.bin/_mocha --compilers js:babel-register
./e2e/*.spec.js`

### DIFF
--- a/e2e/calls.spec.js
+++ b/e2e/calls.spec.js
@@ -1,0 +1,44 @@
+import chai, {
+    expect
+} from 'chai';
+
+var Nexmo = require('../lib/Nexmo');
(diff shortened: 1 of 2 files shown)


e2e/calls.spec.js
```diff
@@ -20,7 +20,7 @@ describe('calls', () => {
     );
   });
 
-  it('should allow a call to be created', () => {
+  it('should allow a call to be created', (done) => {
```
leggetter: @AlexLakatos this should help. Now you'll likely see an error. From here it should be plain sailing ;)

PR #149, opened 2017-10-09
Replace 'var' declarations for 'const' / 'let'

AlvaroBernalG:


mheap:
If we merge this, we'll also need `babel-plugin-transform-es2015-constants` to ensure NodeJS 4 compatibility.

Whilst I think this is a worthwhile change to make, it's a little too risky to merge today until we have better end to end test coverage

kellyjandrews:
@mheap is this something we should still update here, or can we close this?

kellyjandrews:
Going to close this for now.

PR #153, opened 2017-10-14
Added documentation about updating calls

AverageMarcus:
The library supports the ability to update existing calls but there is no mention of it in the docs. Added a small example as well as listing it on the coverage checklist.

### DIFF
--- a/README.md
+++ b/README.md
@@ -116,6 +116,14 @@ nexmo.calls.get(callId, callback);
 
 For more information see https://docs.nexmo.com/voice/voice-api/api-reference#call_create
 
+### Update a Call
+
+```js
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -116,6 +116,14 @@ nexmo.calls.get(callId, callback);
 
 For more information see https://docs.nexmo.com/voice/voice-api/api-reference#call_create
 
+### Update a Call
```
AlexLakatos: Thanks for this @AverageMarcus 🙇 ! I have a small nit pick, I'd move that after "Query calls"

AlexLakatos:
LGTM 🤙 Thanks @AverageMarcus

PR #154, opened 2017-10-14
Handle 204 response

AverageMarcus:
When updating a call a 204 response is returned (such as when hanging up an active call). Currently this is seen as an error by the lib but the action does perform correctly but doesn't provide any response body.


leggetter:
Thanks for this, @AverageMarcus!

A couple of things:

1. The 204 response is a bug that we're waiting to be fixed cc @judy2k 
2. We should handle this anyway, but it'd be good to have a test to check this too

@AlexLakatos - can you please speak to @judy2k about how we're going to handle this until the API bug is fixed?

AverageMarcus:
Well... I made a mess of this PR. Going to re-create and push again with the test added.

PR #155, opened 2017-10-16
Handle 204 status code

AverageMarcus:
When updating a call a 204 response is returned (such as when hanging up an active call). Currently this is seen as an error by the lib but the action does perform correctly but doesn't provide any response body.

(Created again after messing up #154.)

### DIFF
--- a/src/HttpClient.js
+++ b/src/HttpClient.js
@@ -123,6 +123,8 @@ class HttpClient {
           headers['retry-after'] = retryAfterMillis;
         }
         error = {body: data.join('')};
+      } else if (status === 204) {
+        response = data;
       } else if (status >= 400 || status < 200) {
         error = {body: JSON.parse(data.join('')), headers};
       } else if(method !== 'DELETE') {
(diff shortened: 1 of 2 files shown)


AlexLakatos:
Thanks @AverageMarcus ! Continuing the discussion from #154 ,agreed with @judy2k that a `204` should return null.

src/HttpClient.js
```diff
@@ -123,6 +123,8 @@ class HttpClient {
           headers['retry-after'] = retryAfterMillis;
         }
         error = {body: data.join('')};
+      } else if (status === 204) {
+        response = data;
```
AlexLakatos: can we return `null` here instead please?

test/HttpClient-test.js
```diff
@@ -271,6 +271,14 @@ describe('parseResponse', function() {
     expect(callback).was.calledWith({ statusCode: 404, body: { 'error' : 'error' }, headers: headers}, null);
   });
 
+  it ('should parse a 204 status code as a JSON object', function() {
```
AlexLakatos: and update the test as well. We shouldn't parse a `204` as JSON

AlexLakatos (changes requested):

AlexLakatos (approved):

mheap:
LGTM - I've merged it in to master. Thanks @AverageMarcus!

PR #170, opened 2017-12-14
Remove the requirement to specify an API key + secret

### LINKED ISSUE #159: Make API_KEY and API_SECRET optional
When making a voice call, you only need the application ID + private key. The API key and secret aren't used at all.

We should be able to use the library without providing a key/secret if an ppplication ID + private key are provided

mheap:
Initially the plan was to only make it optional if a private key was
provided, but it's tough to do this without significant refactoring at
this time.

As the API responds nicely with a 401 and we handle errors in the lib,
removing this check seems OK to me. We can add the enforcing of
key/secret or private key at a later date

Resolves #159


AlexLakatos (changes requested):
LGTM 🍏 
Please also update the readme saying these are optional for voice applications.

AlexLakatos (approved):

PR #172, opened 2017-12-15
Move account related methods from singleton to Account class

mheap:
There's a ton of test refactoring in here too. It's added a new set of Chai matchers (NexmoChai), a single place to configure Chai (NexmoTestUtils) and a place to configure repetative configs e.g. Credentials.

### DIFF
--- a/src/Account.js
+++ b/src/Account.js
@@ -30,25 +30,29 @@ class Account {
     return this.options.rest.get("/account/get-balance", callback);
   }
 
-  /**
-   * TODO: document
-   */
-  updatePassword() {
+  updatePassword(newSecret, callback) {
+    return this.options.rest.postUseQueryString(
+      "/account/settings",
+      { newSecret },
+      callback
+    );
     this._nexmo.changePassword.apply(this._nexmo, arguments);
   }
 
-  /**
-   * TODO: document
-   */
-  updateSMSCallback() {
-    this._nexmo.changeMoCallbackUrl.apply(this._nexmo, arguments);
+  updateSMSCallback(moCallBackUrl, callback) {
+    return this.options.rest.postUseQueryString(
+      "/account/settings",
+      { moCallBackUrl },
+      callback
+    );
   }
 
-  /**
-   * TODO: document
-   */
-  updateDeliveryReceiptCallback() {
-    this._nexmo.changeDrCallbackUrl.apply(this._nexmo, arguments);
+  updateDeliveryReceiptCallback(drCallBackUrl, callback) {
+    return this.options.rest.postUseQueryString(
+      "/account/settings",
+      { drCallBackUrl },
+      callback
+    );
   }
 
   topUp(trx, callback) {
--- a/src/Account.js
+++ b/src/Account.js
@@ -26,8 +26,8 @@ class Account {
   /**
    * TODO: document
    */
-  checkBalance() {
-    this._nexmo.checkBalance.apply(this._nexmo, arguments);
+  checkBalance(callback) {
+    return this.options.rest.get("/account/get-balance", callback);
   }
 
   /**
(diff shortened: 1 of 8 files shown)


AlexLakatos (changes requested):
LGTM 👨‍🎤 
The only thing I'm not really happy about is having the paths in the methods directly. Having a property on the class that you can then pass to the method would make it easier to update when the API changes, something like the CallResource: https://github.com/Nexmo/nexmo-node/blob/master/src/CallsResource.js#L14

mheap:
I'm not convinced on the property method myself. There aren't enough endpoints + they don't change frequently enough to warrant the extra layer of indirection for someone reading the code IMO

src/HttpClient.js
```diff
@@ -180,6 +180,13 @@ class HttpClient {
   }
 
   get(path, params, callback) {
+    if (!callback) {
```
leggetter, line 183: Note: I think this is internal only, but just wanted to flag in case anything was using this and passing `null` or `{}` for the params.

src/index.js
```diff
@@ -470,24 +465,6 @@ exports.deleteApplication = function(appId, callback) {
   }
 };
 
-exports.changePassword = function(newSecret, callback) {
```
leggetter: Due to this change, I think we should make it a major version bump.

leggetter (commented):
"accounts/settings" is used around 6 times (class and tests) so it may be nice to make this accessible via `Account.SETTINGS_PATH`. If we did that we'd likely also want to shift the pricing one.

But it's not a deal breaker.

src/index.js
mheap: There shouldn't be any public API change due to this. It's still accessible as `nexmo.account.changePassword()`

src/HttpClient.js
mheap: If someone passes `null` or `{}` for params, it'll behave the same way. What issue do you see?

src/index.js
leggetter: Yes, it's accessible via: 

```js
const Nexmo = require('nexmo');
const nexmo = new Nexmo(...);

nexmo.account.changePassword()
```

But it used to be accessible via the singleton:

```js
const nexmo = require('nexmo/lib/index.js')

nexmo.initialize(...);

nexmo.changePassword()
```

That (and others) have been removed so any code that does this (for whatever reason) would now break.

My feeling is that to be safe we should consider a major bump once we move everything out of the singleton.

mheap: Ah ok, I didn't realise `nexmo.changePassword()` was valid (although it was possible). We can bump the major version for this

leggetter: Yeah, anything exported from `index.js` is potentially accessible via the API.

src/Account.js
```diff
@@ -26,29 +26,33 @@ class Account {
...
-   * TODO: document
-   */
-  updatePassword() {
+  updatePassword(newSecret, callback) {
+    return this.options.rest.postUseQueryString(
+      "/account/settings",
+      { newSecret },
+      callback
+    );
     this._nexmo.changePassword.apply(this._nexmo, arguments);
```
AlexLakatos: shouldn't this be deleted as well?

src/index.js
AlexLakatos: Maybe we should keep `index.js` backwards compatible for now with `exports.methodName = account.newMethod` and just remove the exports when we're done with the singleton and do a major version bump.

src/Account.js
mheap: Yep, it should be

src/index.js
mheap: Sounds like a sensible path

mheap:
@AlexLakatos I'm going to merge this to get the NexmoChai changes in to the main branch. We can revisit `PATH()` at DevRelWeek when we put together a plan for the library

PR #174, opened 2017-12-20
Initial pass at media API implementation

mheap:
I've added `request` to handle the multipart form data for us. This is the first step towards using `request` as our main HTTP client

  * Media._makeRequest is required as Content-Type must be application/json
    for all requests
  * I've added a flag (`skipJsonDecoding`) to HTTPClient to allow us to
    download a raw response (required for `media.download()`
    out the mocks for `request`
  * `media.download` implements both `If-Modified-Since` and
    `If-None-Match` - https://ea.developer.nexmo.com/api/media#download-media-file - but it is not documented. This is functionality we'll only expose if
    people explicitly ask for it.
  * Add support for mocking `request` when posting a file.  We check specifically
    for the `postFile` method for now as that's the only thing that uses `request`.
    This will need refactoring if we move away from HTTPClient in the future
  * Add examples for Media API + supported list to README

### DIFF
--- a/src/HttpClient.js
+++ b/src/HttpClient.js
@@ -1,5 +1,6 @@
 var https = require("https");
 var http = require("http");
+var request = require("request");
 var querystring = require("querystring");
 
 class HttpClient {
@@ -15,13 +16,14 @@ class HttpClient {
     };
     this.logger = options.logger;
     this.timeout = options.timeout;
+    this.requestLib = request;
 
     if (options.userAgent) {
       this.headers["User-Agent"] = options.userAgent;
     }
   }
 
-  request(endpoint, method, callback) {
+  request(endpoint, method, callback, skipJsonParsing = false) {
     if (typeof method === "function") {
       callback = method;
       endpoint.method = endpoint.method || "GET";
@@ -92,7 +94,8 @@ class HttpClient {
             response,
             responseData,
             endpoint.method,
-            callback
+            callback,
+            skipJsonParsing
           );
         }
       });
@@ -111,7 +114,7 @@ class HttpClient {
     });
   }
 
-  __parseResponse(httpResponse, data, method, callback) {
+  __parseResponse(httpResponse, data, method, callback, skipJsonParsing) {
     const isArrayOrBuffer = data instanceof Array || data instanceof Buffer;
     if (!isArrayOrBuffer) {
       throw new Error("data should be of type Array or Buffer");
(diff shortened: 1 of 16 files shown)


src/HttpClient.js
```diff
@@ -138,7 +140,11 @@ class HttpClient {
       } else if (status >= 400 || status < 200) {
         error = { body: JSON.parse(data.join("")), headers };
       } else if (method !== "DELETE") {
-        response = JSON.parse(data.join(""));
+        if (!skipJsonParsing) {
```
AlexLakatos: cause of the way js does `!undefined` can we change this please to:
```
if (!!skipJsonParsing) {
  response = data.join("");
} else {
  response = JSON.parse(data.join(""));
}

File: src/HttpClient.js [hunk 2]
```diff
@@ -20,7 +21,7 @@ class HttpClient {
     }
   }
 
-  request(endpoint, method, callback) {
+  request(endpoint, method, callback, skipJsonParsing) {
```
AlexLakatos: can we default `skipJsonParsing` to false here please?
`request(endpoint, method, callback, skipJsonParsing=false) {`

src/HttpClient.js
```diff
@@ -179,7 +185,7 @@ class HttpClient {
     };
   }
 
-  get(path, params, callback) {
+  get(path, params, callback, useJwt) {
```
AlexLakatos: can we default `useJwt` here to false please?

src/Media.js
```diff
@@ -0,0 +1,85 @@
...
+    if (!opts.file && !opts.url) {
+      throw new Error(
+        "You must provide either 'file' or 'url' to upload a file"
+      );
+    }
+
+    if (opts.file) {
+      opts.file = fs.createReadStream(opts.file);
+    }
+    return this.options.api.postFile("/v3/media", opts, callback, true);
```
AlexLakatos: there are 14 instances of `/v3/media` in these files. can we please move it to a class constant for easier maintenance?

mheap:
@AlexLakatos All comments addressed, skipped tests fixed and uncommented and examples added. Could you take another look?

foo.js
```diff
@@ -0,0 +1,82 @@
+var Nexmo = require("./lib/Nexmo");
+
```
AlexLakatos: Is `foo.js` supposed to be in here? And pretty sure that's your API key in there...

AlexLakatos (approved):
LGTM 🎖

PR #176, opened 2018-01-10
Add support for custom timeout setting

holm:
This adds support for adding a custom timeout value. Right now the default Node timeout of 120,000ms is used, which for our needs is too long. 

The default is now 30,000ms, which is more reasonable for a use case like this in my opinion, but feel free to change this.

### DIFF
--- a/src/HttpClient.js
+++ b/src/HttpClient.js
@@ -14,6 +14,7 @@ class HttpClient {
       Accept: "application/json"
     };
     this.logger = options.logger;
+    this.timeout = options.timeout || 30 * 1000;
 
     if (options.userAgent) {
       this.headers["User-Agent"] = options.userAgent;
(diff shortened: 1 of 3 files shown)


mheap:
@holm This looks good to me. Changing the default time out may break things for other consumers so if you could leave the default but add the ability to set a custom timeout that'd be awesome

holm:
Sure, I removed the default value

src/HttpClient.js
```diff
@@ -56,7 +61,7 @@ class HttpClient {
     if (options.port === 443) {
       request = this.https.request(options);
     } else {
-      request = http.request(options);
+      request = this.http.request(options);
```
holm, line 64: A note that I also changed this to use http on the instance. It seems this was the intention.

PR #182, opened 2018-03-20
[WIP] Refactor from singleton

mheap:
I need to test every endpoint and make sure that the tests are representative of real API requests (I did this work on the plane), but the old "everything in `index.js`" way of life is now gone!

I've tried to keep everything as similar to v1 as possible, but I'm sure some backwards compatibility breaks have snuck through. As this removes methods from the `nexmo` object it breaks BC anyway so I'm not too worried, but I feel as though this should form the base of `nexmo-node v2.0.0` (branch name `next`) and provide the opportunities to make some changes that we've been meaning to make for a while.


kellyjandrews:
@mheap - should we implement any of this here, or let v3 handle this change?

kellyjandrews:
Closing for now - if we need to reopen later we can, but I would assume this will be done in v3.

PR #202, opened 2018-09-20
Secret rotation endpoints working

judy2k:
Please can we get a couple of reviews to check whether this is ready to merge?


src/HttpClient.js
```diff
@@ -211,20 +211,28 @@ class HttpClient {
...
-  delete(path, callback, useJwt) {
+  delete(path, callback, useJwt, useBasicAuth) {
     let params = {};
-    if (!useJwt) {
+    if (!useJwt && !useBasicAuth) {
       params["api_key"] = this.credentials.apiKey;
       params["api_secret"] = this.credentials.apiSecret;
     }
 
+      let headers = {};
```
AlexLakatos: indentation looks broken here

src/HttpClient.js
```diff
@@ -311,11 +319,16 @@ class HttpClient {
 
     path = path + joinChar + querystring.stringify(qs);
 
+      let headers = { "Content-Type": "application/json" };
```
AlexLakatos: indentation looks broken here

src/Account.js
```diff
@@ -61,6 +61,22 @@ class Account {
       callback
     );
   }
+
+  listSecrets(account, callback) {
+    return this.options.api.get("/accounts/"+account+"/secrets", {}, callback, false, true);
```
AlexLakatos: does this endpoint support filtering?

src/Account.js
```diff
@@ -61,6 +61,22 @@ class Account {
...
+  }
+
+  getSecret(account, id, callback) {
```
AlexLakatos: the way we handle get/list for the newer parts of the library, we combine into a single method `getSomething` that supports string or empty object params for /id or /?query=string

AlexLakatos: If the endpoint doesn't support filtering, I'd still want to go with a single get method that gets passed an optional string for id and does either single item get or multiple item get

AlexLakatos: the get/list combined method didn't look quite right, abandoned the idea

README.md
```diff
@@ -348,6 +348,30 @@ For more information see https://docs.nexmo.com/tools/application-api/api-refere
 nexmo.account.checkBalance(callback);
 ```
 
+### List Account Secrets
+
+```js
+nexmo.account.listSecrets(account, callback);
```
judy2k: I think all the other libraries use the preferred parameter name `api_key` instead of `account` or `account_id`.

judy2k (commented):
Looks great, but I think the (totally consistent) use of `account` breaks with our documentation policy of referring to this as `api_key`.

TBH I think account is the more accurate term, but I think for consistency's sake, this should probably be changed.

README.md
```diff
@@ -348,6 +348,30 @@ For more information see https://docs.nexmo.com/tools/application-api/api-refere
...
+### List Account Secrets
+
+```js
+nexmo.account.listSecrets(api_key, callback);
+```
+
+### Get Account Secret
+
+```js
+nexmo.account.getSecret(api_key, secretId, callback);
```
judy2k: Now, I realise this is my fault, but I think the naming convention for this library should make it `apiKey` not `api_key`?

judy2k:
Okay, I approve! You are a :star: !

PR #209, opened 2018-10-10
Update year in LICENSE.txt

### LINKED ISSUE #208: Update year in license
Currently, the year in license is 2016.
Would be great if we could change it to 2016-present as this would avoid copyright issues in the future.

rahulkumaran:
Fixes #208 
The year has been updated to 2016-present
This is a one time change and need not be changed every year. This will also ensure no copyright issues arise in the future.
Please do review.
Thanks!

### DIFF
--- a/LICENSE.txt
+++ b/LICENSE.txt
@@ -1,7 +1,7 @@
 
 The MIT License (MIT)
 
-Copyright (c) 2016-present Nexmo Inc
+Copyright (c) 2018 Nexmo Inc
 
 Permission is hereby granted, free of charge, to any person
 obtaining a copy of this software and associated documentation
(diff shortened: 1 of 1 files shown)


AlexLakatos:
The MIT license is not flexible enough to allow perpetuity. Current year needs to be used. Can you update that to say 2018 please?

rahulkumaran:
Done @AlexLakatos

AlexLakatos:
Thank you! Can I ask you do that for other repos as well in http://github.com/nexmo-community, pease?

rahulkumaran:
@AlexLakatos in all the 107 repositories? Would it be possible to get a collaborator access to directly push changes into the LICENSE file? Because otherwise I'll have to fork each of them right?

PR #210, opened 2018-10-12
docs(README): replace var with const/let

TheAlexLichter:
The examples should reflect a "*modern*" code style, which includes `const/let` over `var` ☺️


AlexLakatos:
Thanks for the PR!

TheAlexLichter:
@AlexLakatos You are welcome ☺️

Issue #196, opened 2018-07-21
HTTP request body truncated


mheap:
When making a `PUT` request with a body that contains UTF-8 characters, the existing HTTP client doesn't seem to send all of the data.

e.g. Sending `{"text":"¿Cómo te llamas?"}` results in `{"text":"¿Cómo te llamas?` being sent to the server

There are two missing characters at the end, and two non-ascii characters in the payload (`¿` and `ó`) which leads me to believe that it's encoding related

ferk6a:
Indeed, it is a real bug. I just checked out the source code and found it:
https://github.com/Nexmo/nexmo-node/blob/5a181611b327d730e163cd5f55c77a25c6cfe675/src/StreamResource.js#L30-L39

`Content-Length` is wrong, it's being counted by a string, in "JavaScript codepoints" (a bit weird, not exactly UTF-8 because of surrogate pairs). I will fill a pull request to fix it tomorrow.

PR #212, opened 2018-10-16
Count Content-Length in actual bytes

ferk6a:
Hello! I saw the issue #196 and immediately went to look into the code to figure out why it was happening. Found out it was an issue with UTF-8 data, so I fixed it using the [node documentation examples on how to use the `http` module](https://nodejs.org/docs/latest-v9.x/api/http.html#http_response_writehead_statuscode_statusmessage_headers):

```javascript
const body = 'hello world';
response.writeHead(200, {
  'Content-Length': Buffer.byteLength(body),
  'Content-Type': 'text/plain' });
```

Using `byteLength` rather than string `length` seems to fix the problem, since it was counting characters wrong for UTF-8 encoded data.

I can provide tests for this, but I don't know where exactly they would fit in since there are multiple places in the code where the change has been made.

### DIFF
--- a/src/CallsResource.js
+++ b/src/CallsResource.js
@@ -57,7 +57,7 @@ class CallsResource {
       body: params,
       headers: {
         "Content-Type": "application/json",
-        "Content-Length": params.length,
+        "Content-Length": Buffer.byteLength(params),
         Authorization: `Bearer ${this.creds.generateJwt()}`
       }
     };
@@ -115,7 +115,7 @@ class CallsResource {
       body: params,
       headers: {
         "Content-Type": "application/json",
-        "Content-Length": params.length,
+        "Content-Length": Buffer.byteLength(params),
         Authorization: `Bearer ${this.creds.generateJwt()}`
       }
     };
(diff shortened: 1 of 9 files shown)


mheap:
@fer22f Thanks for this. It looks like we need a test adding for each updated instance, as we can't write a single test due to the code being duplicated.

Each test uses a mocked HTTPClient instance, so it should be reasonable straightforward to assert that the headers provided match what we're expecting (specifically `Content-Length`)

ferk6a:
@mheap I added some UTF-8 tests, some a bit silly (I don't think anyone should use Unicode in URLs directly like that, punycode would probably be used instead), see if you like it :)

The ones in TalkResource are probably the most important.

mheap:
Looks great to me. @AlexLakatos could we get a review?

AlexLakatos (approved):
Looks really great, thanks for the PR!

AlexLakatos:
We're celebrating Hacktoberfest this month, don't forget to fill out the form if you want to get the Nextoberfest T-Shirt https://www.nexmo.com/blog/2018/10/09/join-nextoberfest-tis-the-season-to-be-hacking-dr/

PR #217, opened 2018-10-18
Initial pass at media API implementation

UddhavNavneeth:
There's a lot of room for improvement, but there's enough here to at
least try out the API.

* I've added `request` to handle the multipart form data for us. This
is the first step towards using `request` as our main HTTP client
* Media._makeRequest is required as Content-Type must be application/json
for all requests
* I've added a flag (`skipJsonDecoding`) to HTTPClient to allow us to
download a raw response (required for `media.download()`
out the mocks for `request`
* `media.download` implements both `If-Modified-Since` and
`If-None-Match` - https://ea.developer.nexmo.com/api/media#download-media-file
- but it is not documented. This is functionality we'll only expose if
people explicitly ask for it.
* Add support for mocking `request` when posting a file.  We check specifically
for the `postFile` method for now as that's the only thing that uses `request`.
This will need refactoring if we move away from HTTPClient in the future
* Add examples for Media API + supported list to README

### DIFF
--- a/README.md
+++ b/README.md
@@ -560,6 +560,13 @@ Also see the [Nexmo Node Quickstarts repo](https://github.com/nexmo-community/ne
     * [x] Buy
     * [x] Cancel
     * [x] Update
+* Media
+  * [x] Upload
+  * [x] Download
+  * [x] Search
+  * [x] Get
+  * [x] Update
+  * [x] Delete
 * Voice (Deprecated)
   * [x] Outbound Calls
   * [ ] Inbound Call Webhook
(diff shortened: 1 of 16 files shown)


UddhavNavneeth:
just for hacktoberfest

mheap:
I thought this looked familiar. This isn’t ready to merge.

Please don’t raise PRs with other people’s work. If you’d like to contribute, take a look for opportunities to improve the library.

UddhavNavneeth:
I'm really sorry, I was just trying to do a basic PR for hacktoberfest, I
don't know how to use Github that well.

PR #221, opened 2018-10-23
Add /dev/null equivalent for Windows

### LINKED ISSUE #213: Media#upload Test Fails on Windows
The following fails on Windows because there is no `/dev/null` file:

```js
// From Media-test.js
  describe("#upload", function() {
    // @TODO Add assertions for POST body
    // @TODO Add mocks for request
    it("should call the correct URL (file provided)", function() {
      return expect(this.media)
        .method("upload")
        .withParams({ file: "/dev/null" })
        .to.postFile.to.url("/v3/media");
    });
```

Since the test is only using `/dev/null` as a dummy file to test the upload functionality, it might be worthwhile to either provide a test asset, or to create and destroy a test asset before the test is run.

```
  1) Media
       #upload
         should call the correct URL (file provided):
     Uncaught Error: ENOENT: no such file or directory, open 'C:\dev\null'
```

janowsiany:
Fixes #213

### DIFF
--- a/test/Media-test.js
+++ b/test/Media-test.js
@@ -1,4 +1,5 @@
 import Media from "../lib/Media";
+import os from "os";
 import { expect, sinon, TestUtils } from "./NexmoTestUtils";
 
 describe("Media", function() {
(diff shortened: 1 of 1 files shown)


test/Media-test.js
```diff
@@ -62,9 +63,10 @@ describe("Media", function() {
     // @TODO Add assertions for POST body
     // @TODO Add mocks for request
     it("should call the correct URL (file provided)", function() {
+      const file = os.type() === "Windows_NT" ? "\\\\.\\NUL" : "/dev/null";
       return expect(this.media)
         .method("upload")
-        .withParams({ file: "/dev/null" })
+        .withParams({ file })
```
AlexLakatos, line 69: shouldn't this be `{file: file}`?

janowsiany: It is [shorthand property names notation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Object_initializer#New_notations_in_ECMAScript_2015)

AlexLakatos:
Thanks for the PR!

PR #227, opened 2018-10-26
Refactoring tests to es6

Archaeologist03:
Arrow functions
const and let instead of var

I tried to use pretter for formating but seems like it doesn't work even with eslint and beautify uninstalled.

### DIFF
--- a/test/DtmfResource-test.js
+++ b/test/DtmfResource-test.js
@@ -12,19 +12,19 @@ import Credentials from "../lib/Credentials";
 
 chai.use(sinonChai);
 
-var creds = Credentials.parse({
+const creds = Credentials.parse({
   applicationId: "some-id",
   privateKey: path.join(__dirname, "private-test.key")
 });
-var emptyCallback = () => {};
+const emptyCallback = () => {};
 
 describe("DtmfResource", () => {
-  var httpClientStub = null;
-  var dtmf = null;
+  let httpClientStub = null;
+  let dtmf = null;
 
   beforeEach(() => {
     httpClientStub = sinon.createStubInstance(HttpClient);
-    var options = {
+    const options = {
       httpClient: httpClientStub
     };
     dtmf = new DtmfResource(creds, options);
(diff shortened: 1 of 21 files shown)


test/DtmfResource-test.js
```diff
@@ -12,32 +12,32 @@ import Credentials from "../lib/Credentials";
...
   applicationId: "some-id",
   privateKey: path.join(__dirname, "private-test.key")
 });
-var emptyCallback = () => {};
+const emptyCallback = () => {};
 
 describe("DtmfResource", () => {
-  var httpClientStub = null;
-  var dtmf = null;
+  const httpClientStub = null;
```
AlexLakatos: Tests are failing because of this. You've declaring it as a `const`, but assingig `httpClientStub = sinon.createStubInstance(HttpClient);` in line 26. Making this a `let` instead should fix it.

AlexLakatos (changes requested):
This looks good, thanks for updating your PR!

Could I also please ask you to re-base this, it looks bit messy.

Archaeologist03:
Hi, i never have used rebase/vim before, but i think, this somewhat did the job? 
Also changed const to let.
Is this good? 
ps. Thanks for the patience. :)

test/DtmfResource-test.js
```diff
@@ -12,32 +12,32 @@ import Credentials from "../lib/Credentials";
...
   privateKey: path.join(__dirname, "private-test.key")
 });
-var emptyCallback = () => {};
+const emptyCallback = () => {};
 
 describe("DtmfResource", () => {
-  var httpClientStub = null;
-  var dtmf = null;
+  let httpClientStub = null;
+  const dtmf = null;
```
AlexLakatos: this is causing the tests to fail, it's a `const` but it's being assigned to on line 38.

AlexLakatos:
The tests are still failing, there is a new issue in the same file with a different variable. 

Can you please run the tests locally with `npm run test`, it's going to be easier to spot the errors before it goes to travis-ci.

kellyjandrews:
This PR is really outdated now - closing for now.

PR #234, opened 2018-12-19
- Added extra param to Channel and Dispatch

fauna5:
- Added extra param to Channel and Dispatch  opts = {useBasicAuth: true}
- Added param to HttpClient.post() to allow BasicAuth

Sample api usage:
```
const Nexmo = require('nexmo');

const nexmo = new Nexmo({
  apiKey: 'key,
  apiSecret: 'secret',
});

const message = {
  content: {
    type: 'text',
    text: 'Hello from Nexmo',
  },
};

nexmo.channel.send(
  { type: 'sms', number: '01234567890' },
  { type: 'sms', number: 'Nexmo' },
  message,
  (err, data) => { console.log(data.message_uuid); },
  {useBasicAuth: true}
);
```

The one problem with this solution is that if you want to use `util.promisify` to call the API using a promise wrapper it wont work as the last arg is not the callback. However i prefer this option rather than the alternatives:
- breaking everyone using the API by rearranging params
- having a `sendWithBasicAuth` method

### DIFF
--- a/src/HttpClient.js
+++ b/src/HttpClient.js
@@ -290,9 +290,9 @@ class HttpClient {
     );
   }
 
-  post(path, params, callback, useJwt, headers) {
+  post(path, params, callback, useJwt, useBasicAuth, headers = {}) {
     let qs = {};
-    if (!useJwt) {
+    if (!useJwt && !useBasicAuth) {
       qs["api_key"] = this.credentials.apiKey;
       qs["api_secret"] = this.credentials.apiSecret;
     }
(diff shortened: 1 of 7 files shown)


src/HttpClient.js
```diff
@@ -290,9 +290,9 @@ class HttpClient {
     );
   }
 
-  post(path, params, callback, useJwt, headers) {
+  post(path, params, callback, useJwt, useBasicAuth, headers = {}) {
```
mheap, line 293: This will break other endpoints that use `HttpClient.post` as the parameter order has changed

mheap (changes requested):

src/HttpClient.js
fauna5: @mheap no other endpoints use the post( method. It is documented in the README but the `headers` parameter is not mentioned. I think it's safe to modify this internal API.

image

AlexLakatos:
Can you change the PR to be against beta-olympus please and rebase? The beta branch gets force overridden from beta-stitch and beta-olympus every time we release.

src/HttpClient.js
AlexLakatos: Can you please update the Readme as well with the correct signature?

mheap (approved):

PR #237, opened 2019-01-08
Fix Number Insight  Advanced Async

sammachin:
The Async calls to number insight were not working as per the building blocks, they were making the Synchronous call and there was never a callback to the URL.
It looks as if the naming was mixed up betweeen Sync and Async.
This fixes that issue.
Also we were printing a console warn that in 2.0 advanced would move to being Syncronous, we're now way past that so I've removed the warning to cleanup that block of code.

### DIFF
--- a/src/NumberInsight.js
+++ b/src/NumberInsight.js
@@ -74,15 +74,10 @@ class NumberInsight {
     // remove 'level' as it's a library-only parameter
     delete options.level;
 
-    if (level === "advanced" || level === "advancedAsync") {
-      if (level === "advanced") {
-        console.warn(
-          'DEPRECATION WARNING: Number Insight Advanced with a level of "advanced" will be synchronous in v2.0+. Consider using the level "advancedAsync" to keep using the async option.'
-        );
-      }
-      this._nexmo.numberInsightAdvancedAsync.apply(this._nexmo, arguments);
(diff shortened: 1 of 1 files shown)


garann:
Though I'm the furthest thing from an expert in the subtleties and defaults of the three, it seems like this is a pretty clear and sensible one-to-one mapping for Advanced, Async, and Standard. I'd say :shipit:

garann (approved):

mheap:
This would be a breaking change to the library. Although the comment + building block are misleading we need to keep the existing behaviour as the change wasn't made at 2.0.0.

The building block was updated to be accurate in https://github.com/nexmo-community/nexmo-node-quickstart/commit/2e048f088512f37e0fcc6feec79930b64a615c1d but wasn't pulled in to NDP, I'll resolve that now

sammachin:
~So in order to make an _Async_ request you should set the level to `advancedSync` ??? That makes zero sense!~

Ah no I see so to do an Async Call you can set the level to `advanced` (for legacy compatibility) or `advancedAsync`, 
To make the (new) Synchronous Call you set it to `advancedSync` 

Confusing product naming!

sammachin:
Perhaps add a comment to that block of code referencing this thread so that someone else doesn't try to "fix" it in the future

Issue #190, opened 2018-06-11
TypeScript - Typings


ghost:
Hi,

I am currently using your NPM module in a TypeScript project.

I tried to install the typings using the following commands `npm i --save-dev @types/nexmo` and `npm i --save-dev @types/nexmo-node`. Both commands end up with a 404.

Are you planning to provide a `.d.ts` file for your NPM module?

AlexLakatos:
Hi @jackTheRipper ,
We are not planning to support TypeScript in the near future. Could you tell us about your use case so we can consider it?

Cheers,
Alex

ghost:
Hi @AlexLakatos,

I am currently building a micro-service to interact with your APIs. The micro-service is using your NPM module and is entirely build in TypeScript.

For now, I can compile by creating an `index.d.ts` with this single line inside `declare module 'nexmo';`.

It works but I am loosing all the advantages of working with TypeScript.

Could it be possible for you to plan an `index.d.ts` with all the signature of your modules and functions?

ghost:
Hi @AlexLakatos , any update for that?

AlexLakatos:
Hi @jackTheRipper,
I've considered adding TypeScript typings and won't have the time until later this year with the re-write in version 3. Let me know if you'd like to add the typings yourself and I'll try to help.

Cheers,
Alex

ghost:
Hi @AlexLakatos, that’s ok for me. However I will push it per functionalities if it’s ok for you instead of opening a big PR which will take ages for being reviewed.

I will start with the “verify” functionality. Is it ok for you?

AlexLakatos:
Perfect! Let me know if you need any help. If you want to chat in sort of real time, please join our community slack: https://developer.nexmo.com/community/slack

julien-c:
Very interested in this too. Considering the (small) size of the codebase, a full rewrite to Typescript would be even better (and seems like not too big of a project).

Let me know if that is something you'd be interested in.

ghost:
Hey, just finished the verify API for our project. I will push the .d.ts
file this week end

ghost:
Done.

julien-c:
Awesome @jackTheRipper

AdelBachene:
@jackTheRipper @AlexLakatos  any Update on publishing the .d.ts ?

ghost:
Hi @AdelBachene - I updated the PR based on the comment of @AlexLakatos - let's see if everything is alright for him

[...]

PR #241, opened 2019-02-02
Add types for verify API

ghost:
As discussed in the issue #190 , here are the typings for the Verify API.

It follows the specs of the Rest API described in your documentation: https://developer.nexmo.com/api/verify

The `package.json` was updated as well for targeting the types.

### DIFF
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -4,6 +4,7 @@ declare module 'nexmo' {
         apiKey: string;
         apiSecret: string;
         applicationId: string;
+        privateKey?: string;
     }
 
     export interface NexmoApiError {
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -1,0 +1,106 @@
+declare module 'nexmo' {
+    /* general */
+    export interface CredentialsObject {
+        apiKey: string;
+        apiSecret: string;
+        applicationId: string;
+    }
+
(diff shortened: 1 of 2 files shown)


typings/index.d.ts
```diff
@@ -0,0 +1,106 @@
+declare module 'nexmo' {
+    /* general */
+    export interface CredentialsObject {
+        apiKey: string;
+        apiSecret: string;
+        applicationId: string;
```
AlexLakatos, line 6: the CredentialsObject should have a privateKey in there as well

ghost: oki, will add it

ghost:
I noticed I did a mistake as well, the errors status should be string not number - fixing it

typings/index.d.ts
ghost: fixed

AlexLakatos:
Thanks, these look good. I'll merge and release a new version after Easter, next week.

felipeplets (approved):

jeremyhon:
hi, is there any progress on this?

AlexLakatos:
Releasing a new version with this today, thanks for the patch!

AlexLakatos:
just released nexmo@2.5.1 with Verify typings, and added the workflow_id parameter for the Request as well.

PR #246, opened 2019-04-01
Correcting note about opts in readme

S33G:


AlexLakatos:
thanks for sending the PR

PR #247, opened 2019-04-01
Showing which constructor params are optional

S33G:
I was under the impression that all of these values were required at first.


AlexLakatos:
Thanks for the PR!

PR #251, opened 2019-05-10
Update Numbers API examples in README

marklewin:
- New API reference doc links
- Missing `type` parameter (`mobile-lvn`, `landline` etc) in search example


AlexLakatos (approved):
LGTM 🔩

PR #263, opened 2019-09-09
Add simple contributing file

lornajane:
We didn't have a contributing file but we do like contributions, so I added one!


AlexLakatos (approved):
LGTM 🏆 . Thank you @lornajane !

PR #264, opened 2019-09-09
Bump eslint version due to security warning

lornajane:
I bumped eslint to avoid a regex attack vulnerability


AlexLakatos (approved):
LGTM 📧

PR #265, opened 2019-09-10
Add Missing Conversation, Member and Event endpoints

AlexLakatos:


src/Events.js
```diff
@@ -0,0 +1,94 @@
...
+   * Get an existing event.
+   *
+   * @param {string} [conversationId] - The unique identifier for the conversation
+   * @param {string|object} query - The unique identifier for the event to retrieve
+   *               or a set of filter parameters for the query. For more information
+   *               see https://developer.nexmo.com/api/conversation#getEvents
+   * @param {function} callback - function to be called when the request completes.
+   */
+  get(conversationId, query, callback) {
+    this._nexmo.getWithQuery(
```
leggetter: Can we please update all of these GET requests to not use the singleton? The custom-events branch shows how to do this and it avoids the problems we've seen in using the singleton.

leggetter (changes requested):
I would request that we move away from using `getWithQuery` on the singleton everywhere. We certainly shouldn't be introducing more uses of it. This is my only requested change.

Comments:

I found https://github.com/Nexmo/nexmo-node/blob/custom-events/examples/ex-create-conversation-event.js very useful during the development process as it removed a lot of the setup. But maybe that's just something that is throwaway after the initial development.

Finally, is the `query` on `Events.get` optional? It feels like it should be unless that inconsistent with how we've approached `get` on other objects. We should be consistent.

src/Events.js
leggetter: https://github.com/Nexmo/nexmo-node/blob/custom-events/src/Events.js#L60-L75

src/Conversations.js
```diff
@@ -95,6 +101,30 @@ class Conversations {
     this.options.httpClient.request(config, callback);
   }
 
+  /**
+   * Record an existing conversation.
+   *
+   * @param {string} [conversationId] - The unique identifier for the conversation to record.
```
leggetter: What do the square brackets around the `conversationId` mean in the JSDoc?

AlexLakatos: they mean optional, but the conversationId is not optional here, good catch

src/Events.js
AlexLakatos: `(typeof eventId !== "function" ? `/${eventId}` : "")` won't work with query. I'll move the getWithQuery logic in a utils class so we move away from the singleton. Would that be ok with you?

AlexLakatos:
> Finally, is the `query` on `Events.get` optional? It feels like it should be unless that inconsistent with how we've approached `get` on other objects. We should be consistent.

It's not optional. The Conversation endpoints were modeled after Calls when I started, and query isn't optional there. So for consistency, we'll keep it non-optional until the re-write.

src/Events.js
leggetter: Yeah 👍

PR #271, opened 2019-10-18
Conversations API Beta 2 Updates

AlexLakatos:
- add beta2 API version for list endpoints
- add cursor pagination methods next() and prev() for conversations, users, members and events
- add v1 API version for record endpoint
- add the ability to change hosts for the Conversations API
- add query parameter for nexmo.users.getConversations


README.md
```diff
@@ -541,18 +579,42 @@ nexmo.users.get(userId, callback);
...
+### Get paginated Users
 
 ```js
 nexmo.users.get(options, callback);
 ```
 
+### Get next page of Users
+
+```js
+nexmo.users.next(users_list_response, callback);
```
dragonmantank: This should probably match the other examples and use `response` instead of `users_list_response`, for consistency.

README.md
```diff
@@ -541,18 +579,42 @@ nexmo.users.get(userId, callback);
...
+### Get paginated Conversations for a User
+
+```js
+nexmo.users.getConversations(userId, options, callback);
+```
+
+### Get next page of Conversations for a User
+
+```js
+nexmo.users.next(conversations_list_response, callback);
```
dragonmantank: Same here, with the change to `response` instead of `conversations_list_response`

examples/ex-record-conversation.js
```diff
@@ -13,9 +13,11 @@ module.exports = function(callback, config) {
 
   nexmo.conversations.record(config.CONVERSATION_ID, {
     action: "start",
-    event_url: "https://example.com/event",
+    event_url: ["https://example.com/event"],
```
dragonmantank, line 16: Is this correct? The spec looks like it's a string, not an array:

https://github.com/Nexmo/api-specification/blob/1654e87ac02f96fec928ec7c7dcb1729638bd3eb/definitions/conversation.yml#L693

examples/run-examples.js
```diff
@@ -75,28 +75,39 @@ exampleFiles = [
...
   // 'ex-get-member.js',
   // 'ex-create-event.js',
   // 'ex-delete-event.js',
   // 'ex-get-event.js',
   // 'ex-get-events.js',
-  // 'ex-create-read-update-delete-conversation.js'
+  // 'ex-get-next-events.js',
+  // 'ex-get-prev-events.js',
+  // 'ex-create-read-update-delete-conversation.js',
+  'ex-create-conversation.js',
```
dragonmantank, line 103: Should this be commented out as well?

dragonmantank (changes requested):
Couple of minor things, otherwise I think this looks good.

examples/ex-record-conversation.js
mheap: The spec is wrong in this case. `event_url` should always be an array of strings (and I've checked, you get a `400` if it's not)

examples/run-examples.js
mheap: It's fine to leave this uncommented. This is a development file and the uncommented line is usually the last thing we tested

PR #272, opened 2019-10-18
Update types to match API

rorhug:
Some small differences i noticed against [the docs](https://developer.nexmo.com/api/verify)

Errors also seem to use string for `status` rather than number as suggested by the `VerifyError` interface and associated enums.
Left that though for feedback.

### DIFF
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -21,7 +21,7 @@ declare module 'nexmo' {
 
     /* verify API */
     export interface VerifyError extends NexmoApiError {
-        status: RequestResponseStatusCode | ControlResponseStatusCode | number;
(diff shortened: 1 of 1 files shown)


rorhug:
Red tests seem to be unrelated to change an are passing on some systems.

AlexLakatos:
Thanks for the PR. I've restarted the failing jobs, they do look intermittent.

In the meantime, you can go ahead and change the status to string as well, that's correct.

rorhug:
Updated the error strings there @AlexLakatos 

Unfortunately [the docs](https://www.typescriptlang.org/docs/handbook/enums.html#string-enums) say string enums "don’t have auto-incrementing behaviour" so had to add them manually.

AlexLakatos:
Thanks for the PR!

PR #274, opened 2019-11-07
Add new numbers fields to README

lornajane:
We added two new fields to the Numbers API but I can use them with this library already since it simply passes the parameters through. Added them to the README for clarity and if there's an issue somewhere for adding `has_application` and `application_id` support to node: I've tested this and it works great :)


AlexLakatos:
Thanks for updating the readme!

AlexLakatos (approved):

PR #282, opened 2019-12-05
Add Conversations v0.1 Support

AlexLakatos:
- Added support for Conversation API v0.1
- Removed `nexmo.users.getConversations`


hummusonrails (commented):
Am I reading this correctly that the primary change here is updating the URL path in the functions?

AlexLakatos:
Yeah, it's a pass-through SDK. So changing the base URL is all that's needed. There's a minor removal of multiple API paths at the same time and removing the users.getConversations method.

hummusonrails (approved):
🚀 _(... but not yet 😁 )_

aghahuseynov (approved):

kellyjandrews:
Does anyone know if any of these changes are still valid after seven months?

dragonmantank:
There will be some changes when v0.3 drops here relatively soon. It's close to this v0.1, but some minor changes.

lornajane:
I believe that these changes were rolled into the beta branch which contains a combined set of changes for conversation, messages, and .... something else. These should have been separate branches IMO, releasing one of the beta featuresets and not another may be tricky with the current setup. This PR is outdated I _think_.

mheap:
This is a separate Conversations implementation using the new spec, not what is currently in the beta branch

PR #283, opened 2020-01-08
encoding query values in signed sms so that SMSs containing more than…

slorello89:
encoding query values in signed sms so that SMSs containing more than 1 word can be signed


AlexLakatos:
Thanks for the PR @slorello89 ! If you could add a test somewhere in there for it as well, that would be awesome. https://github.com/Nexmo/nexmo-node/blob/master/test/HttpClient-test.js#L211|

Bump the package number, and add an entry into the Changelog for it, and we should be good to release this.

AlexLakatos (approved):
LGTM

Issue #278, opened 2019-11-28
Typescript: message proprety does not exist on type 'Nexmo'


soullivaneuh:
With the following code:

```typescript
import Nexmo from 'nexmo';
const nexmo = new Nexmo({
  // @see https://github.com/Nexmo/nexmo-node/issues/277
  applicationId: '',
  apiKey: 'xxx',
  apiSecret: 'xxx',
});
nexmo.message.sendSms('test', '620919191', 'this is a test');
```

I have:

```
TS2339: Property 'message' does not exist on type 'Nexmo'
```

Because of that, the library is unusable at all.

soullivaneuh:
Correction, it works this way:

```typescript
const Nexmo = require('nexmo');
const nexmo = new Nexmo({
  // @see https://github.com/Nexmo/nexmo-node/issues/277
  applicationId: '',
  apiKey: 'xxx',
  apiSecret: 'xxx',
});
nexmo.message.sendSms('test', '620919191', 'this is a test');
```

But now I can take benefit of strict typing. For example with this class:

```typescript
import { Service, NedbServiceOptions } from 'feathers-nedb';
import { Params } from 'express-serve-static-core';
import { Application } from '../../declarations';

const Nexmo = require('nexmo');

interface SmsData {
  _id?: string;
  to: string;
  message: string;
}

export class Sms extends Service<SmsData> {
  nexmo: Nexmo;

  constructor(options: Partial<NedbServiceOptions>, app: Application) {
    super(options);

    this.nexmo = new Nexmo({
      // @see https://github.com/Nexmo/nexmo-node/issues/277
      applicationId: '',
      apiKey: 'xxx',
      apiSecret: 'xxx',
    });
  }

  create(data: SmsData, params?: Params) {
    this.nexmo.message.sendSms('Kidways', data.to, data.message)

    return super.create(data, params);
  }
}
```

I will have the following error on the class property:

```
TSError: ⨯ Unable to compile TypeScript:
src/services/sms/sms.class.ts(14,10): error TS2749: 'Nexmo' refers to a value, but is being used as a type here.
```

I think import method should be fully supported to fix this issue.

AlexLakatos:
The typings support is partial for now and the typings work for the Verify API currently. The SMS API is next on the list to add support for.
We also accept pull requests if you'd like to help fix this issue.

soullivaneuh:
[...]

PR #290, opened 2020-03-02
add some typings for message api

jin60641:
#278 
It will support Message API's sendSms method

reference
[MessageRequestResponseStatusCode](https://developer.nexmo.com/messaging/sms/guides/troubleshooting-sms)
[MessageRequestResponseSuccess, MessageError, MessageRequestResponse, SendSmsOptions, SendSms, Message](https://developer.nexmo.com/api/sms)


AlexLakatos:
Messages API is only supported on the beta branch for now, so I'll switch this on the beta branch for now.

AlexLakatos:
@jin60641 this breaks backwards compatibility for Verify. Can you please switch back the Verify responses and add a TODO note to change them when we release a new major version?

AlexLakatos (changes requested):
Switch back the Verify namespacing.

jin60641:
> @jin60641 this breaks backwards compatibility for Verify. Can you please switch back the Verify responses and add a TODO note to change them when we release a new major version?

I switch back the Verify api's namesapcing and add TODO comment on top of file.
Thanks

typings/index.d.ts
```diff
@@ -19,6 +31,99 @@ declare module 'nexmo' {
         parseError: Error;
     }
 
+    /* message API */
+    export enum MessageRequestResponseStatusCode {
```
Abhi347, line 35: The status codes are returned as string, not integers. You should change the codes to string.

Coooi:
Any updates on this?
I really looking forward to it.

kellyjandrews:
I will evaluate this later this week along with #299 and get these added in.  Thanks for your patience.

typings/index.d.ts
jin60641: I pushed a new commit. check please.

kellyjandrews:
I pulled these in for now and will start a larger change for v2.9.  Thanks for the contribution.  

If you want to help out with others, I will be creating a new branch `typing-updates` to work towards a more complete set of types.

PR #293, opened 2020-03-19
Add host overriding to Beta

AlexLakatos:


lornajane:
I started to review this and I now think we need a wider discussion about hosts vs URLs for overriding so this is now stuck while we talk about that. Sorry :/

lornajane (approved):
There are a lot of changes in this branch which I think is because there are changes from master coming into beta at the same time (maybe those should be in a separate PR in future) but the host override does work for a beta endpoint so I'm approving!

PR #302, opened 2020-06-05
Remove the Singleton

AlexLakatos:


examples/ex-number-buy.js
```diff
@@ -0,0 +1,13 @@
...
+  var Nexmo = require('../lib/Nexmo');
+
+  var nexmo = new Nexmo({
+    apiKey: config.API_KEY,
+    apiSecret: config.API_SECRET
+  }, {
+    debug: config.DEBUG
+  });
+
+  nexmo.number.buy("GB", "447418349929", callback);
```
slorello89: Are we ok with a magic number in buy field (this is problematic over other magic strings which I see e.g. GB, https://example.com - as a number can be unavailable/available depending on availability)

examples/ex-number-cancel.js
```diff
@@ -0,0 +1,13 @@
...
+  var Nexmo = require('../lib/Nexmo');
+
+  var nexmo = new Nexmo({
+    apiKey: config.API_KEY,
+    apiSecret: config.API_SECRET
+  }, {
+    debug: config.DEBUG
+  });
+
+  nexmo.number.cancel("GB", "447418349929", callback);
```
slorello89: magic number again

examples/example.env
```diff
@@ -5,10 +5,19 @@ NEW_API_SECRET=
...
 CODE=
-PRIVATE_KEY=./examples/private.key
-DEBUG=true
+APP_ID=
+CONVERSATION_ID=
+USER_ID=
+MEMBER_ID=
+EVENT_ID=
+SERVICE_MESSAGE_ID=
+PRIVATE_KEY=
```
slorello89: is there some overarching assumption that the `PRIVATE_KEY` is going to be located at `./examples/private.key`? if so it may be worth keeping the pre-set value in the env file.

examples/run-examples.js
```diff
@@ -68,10 +68,18 @@ exampleFiles = [
   // 'ex-get-calls.js',
   // 'ex-make-call.js',
   // 'ex-number-insight-basic.js',
-  // 'ex-send-sms.js',
+  // 'ex-number-insight-standard.js',
+  // 'ex-number-insight-advanced-sync.js',
+  // 'ex-number-insight-advanced-async.js',
+  // 'ex-send-shortcode-alert.js',
+  // 'ex-send-shortcode-2fa.js',
+  'ex-send-sms.js',
```
slorello89: should this be uncommented?

lornajane (approved):
Super difficult to test something this size, but we can't avoid it when it's a structural change! So far:
 - I can't break anything at all
 - The API key and secret are now optional in the constructor which is excellent news for clients that don't need those details
 - I can have multiple client instances, for example to work with different products or with different config settings.

I'll approve but I'd like an example of the ubiquitous `callback` value declared somewhere I can find it ... in the `README` would be ideal. Even if it's just a function that logs what it received.

Can we also include the instruction to `npm run compile` when switching branches or making local changes? It's sort of there for first installation in the CONTRIBUTING file but I feel like I've screwed this up too many times and writing it down could help us in the future.

PR #305, opened 2020-06-11
Add PSD2 Support for Verify

AlexLakatos:
This PR adds `nexmo.verify.psd2` as a method, and updates all the necessary files for release.

To run it:
- switch to the `psd2` branch
- `npm compile`
- uncomment the added examples in `run-examples.js`
- `node run-examples.js`

Note: you can't run both examples at the same time, Verify won't let you run concurrent verifications on the same number. So each at a time. If you don't want to wait 5 minutes and receive an SMS and 2 phone calls in the process, uncomment the `check-verify` example, and run it with the code you got from psd2.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,8 +2,8 @@
 
 All notable changes to this project will be documented in this file. This project adheres to [Semantic Versioning](http://semver.org/).
 
-## 2.7.1
-- ADDED: Support for Verify PSD2 requestes via `nexmo.verify.psd2()`.
+## 2.8.0
+- ADDED: Support for Verify PSD2 requests via `nexmo.verify.psd2()`.
 
 ## 2.7.0
 - ADDED: Made `apiKey` and `apiSecret` optional when `applicationId` and `privateKey` are present in Nexmo constructor.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 All notable changes to this project will be documented in this file. This project adheres to [Semantic Versioning](http://semver.org/).
 
+## 2.7.1
+- ADDED: Support for Verify PSD2 requestes via `nexmo.verify.psd2()`.
+
 ## 2.7.0
 - ADDED: Made `apiKey` and `apiSecret` optional when `applicationId` and `privateKey` are present in Nexmo constructor.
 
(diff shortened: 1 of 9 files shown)


src/Verify.js
```diff
@@ -54,6 +54,24 @@ class Verify {
     }
   }
 
+  /**
+   * TODO: document
```
slorello89, line 58: Needs Summary docs

package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "nexmo",
   "author": "nexmo",
-  "version": "2.7.0",
+  "version": "2.7.1",
```
slorello89: This is a new feature - thus it should be a new minor release

CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 All notable changes to this project will be documented in this file. This project adheres to [Semantic Versioning](http://semver.org/).
 
+## 2.7.1
```
slorello89: should be 2.8.0

slorello89 (changes requested):
Ran everything and it looked good - just a minor doc item and this should be a minor release. Killer last contribution from @AlexLakatos

CHANGELOG.md
lornajane: Why should it be 2.8.0? It's a minor, non-breaking additional change so a patch seems OK to me. If we can clarify, I'll review and try to merge/release

slorello89: As this is a minor non-breaking new feature that we're adding - it would be appropriate to increment the minor version. This is my understanding from [semver](https://semver.org/)

```
Given a version number MAJOR.MINOR.PATCH, increment the:

MAJOR version when you make incompatible API changes,
MINOR version when you add functionality in a backwards compatible manner, and
PATCH version when you make backwards compatible bug fixes.
```

src/Verify.js
lornajane: Nothing else has this so not a blocker for now IMO.

package.json
lornajane: Updated

CHANGELOG.md
lornajane: Thanks! I've switched it to 2.8.0

slorello89 (approved):
LGTM

PR #307, opened 2020-07-07
Treat keys as strings by default

garann:
This change attempts to make supplying a private key more robust. It treats a supplied string as the key itself, instead of as the path to the key, by default. It also converts "\n" to an actual newline and trims the spaces from a "newline-space-space" pattern I kept finding in keys generated via the dashboard.

### DIFF
--- a/src/Credentials.js
+++ b/src/Credentials.js
@@ -39,16 +39,11 @@ class Credentials {
 
     if (privateKey instanceof Buffer) {
       this.privateKey = privateKey;
-    } else if (
-      typeof privateKey === "string" &&
-      privateKey.startsWith("-----BEGIN PRIVATE KEY-----")
-    ) {
-      this.privateKey = new Buffer(privateKey);
-    } else if (privateKey !== undefined) {
-      if (!fs.existsSync(privateKey)) {
-        throw new Error(`File "${privateKey}" not found.`);
-      }
-      this.privateKey = fs.readFileSync(privateKey);
+    } else if (typeof privateKey === "string") {
+      if (fs.existsSync(privateKey)) {
+        privateKey = fs.readFileSync(privateKey);
+      } 
+      this.privateKey = Buffer.from(privateKey.replace(/\\n/g, '\n').replace(/\n\s\s/g, '\n'), 'utf-8');     
     }
 
     /** @private */
--- a/src/Credentials.js
+++ b/src/Credentials.js
@@ -38,12 +38,21 @@ class Credentials {
     this.signatureMethod = signatureMethod;
 
     if (privateKey instanceof Buffer) {
+      // it is already a buffer, use it as-is
       this.privateKey = privateKey;
     } else if (typeof privateKey === "string") {
+      // is this a filename that we can find?
       if (fs.existsSync(privateKey)) {
-        privateKey = fs.readFileSync(privateKey);
-      } 
-      this.privateKey = Buffer.from(privateKey.replace(/\\n/g, '\n').replace(/\n\s\s/g, '\n'), 'utf-8');     
+        // it is a file, use its contents
+        this.privateKey = fs.readFileSync(privateKey);
+      } else {
+        // a string but not a filename, does it look like a key?
+        if(privateKey.startsWith("-----BEGIN PRIVATE KEY-----")) {
+          // OK it's a key. Check for \n, replace with newlines
+          privateKey = privateKey.replace(/\\n/g, '\n');
+          this.privateKey = Buffer.from(privateKey, 'utf-8');
+        }
+      }
     }
 
     /** @private */
--- a/examples/ex-keyfile-creds.js
+++ b/examples/ex-keyfile-creds.js
@@ -1,0 +1,16 @@
+module.exports = function(callback, config) {
+  var Nexmo = require("../lib/Nexmo");
+
+  var nexmo = new Nexmo(
+    {
+      applicationId: config.APP_ID,
+      privateKey: config.PRIVATE_KEY
+    },
+    { debug: config.DEBUG }
+  );
+
+  if(!nexmo.credentials.privateKey) {
+    // no key found
+    throw new Error("PrivateKey not loaded");
+  }
+};
--- a/examples/run-examples.js
+++ b/examples/run-examples.js
@@ -53,6 +53,7 @@ function runExample(exampleFile, callback) {
 // By default all examples are run.
 // Use this array to run a select number of examples.
 exampleFiles = [
+  // 'ex-keyfile-creds.js',
   // 'ex-change-api-host.js',
   // 'ex-change-rest-host.js',
   // 'ex-check-balance.js',


mheap:
> trims the spaces from a "newline-space-space" pattern I kept finding in keys generated via the dashboard

Can you expand on this more? If it's a dashboard issue we should fix it there so that non-SDK users don't hit the issue

garann:
> > trims the spaces from a "newline-space-space" pattern I kept finding in keys generated via the dashboard
> 
> Can you expand on this more? If it's a dashboard issue we should fix it there so that non-SDK users don't hit the issue

Not much I can add, unfortunately. The `private.key` file generated via the dashboard seems to have padding at the end of each line. I noticed it when trying to convert the keys from multiline to one line with escaped newlines to get around dotenv not allowing multiline variables.

mheap:
Do the trailing spaces prevent the key from working, or is this fix a normalising step?

garann:
> Do the trailing spaces prevent the key from working, or is this fix a normalising step?

I'd thought they stopped it from working, but just now trying to find the exact error I can't reproduce it. My thinking was if we're advising doing replacements on the keys as our standard way of dealing with them it might be nice to move that responsibility to the SDK rather than the developer.

mheap:
I agree, but would argue that we should be moving that responsibility further up the chain than implementing it in each SDK.

The description for this PR explains what the change does, but not why the change has been made. Is there anything I can read to get more context?

garann:
> The description for this PR explains what the change does, but not why the change has been made. Is there anything I can read to get more context?

I've mentioned you in a thread on slack to give context. It just seemed like assuming the parameter was a file path wasn't necessarily what you'd be expecting.

lornajane:
I am really pleased we're revisiting this issue, it's a stumbling block! I would prefer if we didn't change the default behaviour though (and I think some of the tests failed as a result of doing so).

I would love to be able to pass in a string and have it recognised as a private key and handle a selection of line ending "features", whether those are coming from the dashboard, from different operating systems, or from some other public/private key tool that a user can be using. Perhaps the first step is to add tests for our desired behaviours.

A side note but can we add this functionality to the main branch and have it cascade out to beta rather than only adding it to beta?

lornajane:
Right. I've extended what Garann started here and added an example so that if we can't load a key, the build/tests (are these even tests?) will fail and we will spot it. Test by setting the various desired supported examples in `examples/.env` and running the `examples/ex-keyfile-creds.js` example. I'm not sure if we need to do something with env vars on the build platform now though ...

If you pass a buffer: we accept it
If you pass a string, we first try to use it as a filename (our previous preferred approach) and read it in and use it. If it wasn't a file we could open, we look if it "looks like" a key (has the correct first line), and we replace any `\n` characters with actual newlines, then use that.
The code didn't fail loudly if it couldn't read a key, so it still does not do that.

@garann I don't have an example of the extra spaces bug you had seen on the dashboard keys. Do we have that somewhere so we can test with it? I think we can harmlessly add another replace for this (maybe before this existing replace?) since if the combination isn't in the key, nothing will happen.

src/Credentials.js
```diff
@@ -38,17 +38,21 @@ class Credentials {
...
+      if (fs.existsSync(privateKey)) {
+        // it is a file, use its contents
+        this.privateKey = fs.readFileSync(privateKey);
+      } else {
+        // a string but not a filename, does it look like a key?
+        if(privateKey.startsWith("-----BEGIN PRIVATE KEY-----")) {
+          // OK it's a key. Check for \n, replace with newlines
+          privateKey = privateKey.replace(/\\n/g, '\n');
+          this.privateKey = Buffer.from(privateKey, 'utf-8');
+        }
```
mheap, line 54: Could we end up in a situation where `this.privateKey` is null if it's a string that doesn't look like a key? If so, let's `throw` early

kellyjandrews: Agree with throwing early. No reason to let it pass if it fails all of these checks.

src/Credentials.js
```diff
@@ -38,12 +38,21 @@ class Credentials {
...
+      // is this a filename that we can find?
       if (fs.existsSync(privateKey)) {
-        privateKey = fs.readFileSync(privateKey);
-      } 
-      this.privateKey = Buffer.from(privateKey.replace(/\\n/g, '\n').replace(/\n\s\s/g, '\n'), 'utf-8');     
+        // it is a file, use its contents
+        this.privateKey = fs.readFileSync(privateKey);
+      } else {
+        // a string but not a filename, does it look like a key?
+        if(privateKey.startsWith("-----BEGIN PRIVATE KEY-----")) {
```
garann, line 50: Of course I didn't record which permutation it was 🤦‍♀️ but this appeared to be failing for me with dashboard-generated keys under some circumstances. Is it strictly needed? If this is the last option, might as well just set whatever's provided as the key, no?

garann:
> @garann I don't have an example of the extra spaces bug you had seen on the dashboard keys. Do we have that somewhere so we can test with it? I think we can harmlessly add another replace for this (maybe before this existing replace?) since if the combination isn't in the key, nothing will happen.

I have several, I could send you one or could just test this locally, what is best? 

I had three `replace`s in there at one point I think and it was just fine. I think you'd only get in trouble if they were ordered in such a way you ended up escaping something that was already escaped, like `\n` -> `\\n` -> `\\\n`

src/Credentials.js
kellyjandrews: I think it's nice to have a sanity check on the key so if it's not matching the pattern it would error immediately - but would be interested why this failed with keys created in the dashboard.  This is all something to evaluate while reworking v3 as well.

kellyjandrews:
Any reason why this couldn't go into the main version as well? It's currently on the beta branch, but I'd think we can add this to the v2.9 release as well.  Let me know if I'm missing something. *Edit* Just saw the same comment from Lorna above :D

lornajane:
+1 to getting it into both main and beta branches

kellyjandrews:
@garann Where are we on this - I'd like to get it merged into sdk this week for a 2.9 release. Let me know if I can help :)

lornajane:
I refactored and there was some discussion of failing more assertively but that seems like a behaviour change to me so I wasn't sure. Also the build is failing so perhaps I could get some advice on what I broke as well? @kellyjandrews I don't see a formal review from you so I'm not sure what you'd like changed before this is ready but let's try to put our heads together on it.

kellyjandrews:
@lornajane I will look at why the tests are failing and go from there - I didn't want to formally review something failing, as I was assuming there was still work to be done.  I'll pull this in and see what's up.

garann:
> @garann Where are we on this - I'd like to get it merged into sdk this week for a 2.9 release. Let me know if I can help :)

I think @lornajane's code replaces most of mine. I wasn't understanding the direction we wanted to go with this, I think.

kellyjandrews:
Thank you both - sorry for not following very well here - I'll get this looked at and wrapped up tomorrow.

kellyjandrews (changes requested):
Based on what I'm seeing, let's leave the error handling alone, as we aren't throwing anything now when null far as I can tell.  The tests seem to be failing due to mocha configuration issues.  

Ideally I'd like this in the main branch as well, so my recommendation is to branch off master, add the code there, and then when i sync up beta and master we will get this in both places.

kellyjandrews:
Closing in favor of #315

PR #310, opened 2020-07-23
Add `target_api_key` option for number.buy

### LINKED ISSUE #296: Extend number.buy() method params
the API has additional param `target_api_key` which need to add to the lib method. Thanks.

kellyjandrews:
Fixes #296

### DIFF
--- a/src/Number.js
+++ b/src/Number.js
@@ -102,20 +102,28 @@ class Number {
   /**
    * TODO: document
    */
-  buy(countryCode, msisdn, callback) {
+  buy(countryCode, msisdn, targetApiKey, callback) {
     if (!countryCode || countryCode.length !== 2) {
       Utils.sendError(callback, new Error(Number.ERROR_MESSAGES.countrycode));
     } else if (!msisdn) {
       Utils.sendError(callback, new Error(Number.ERROR_MESSAGES.msisdn));
     } else {
+      let opts = {
+        country: countryCode,
+        msisdn,
+        target_api_key: targetApiKey,
+        api_key: this.creds.apiKey,
+        api_secret: this.creds.apiSecret
+      };
+
+      if (targetApiKey instanceof Function) {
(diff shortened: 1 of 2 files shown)


src/Number.js
```diff
@@ -102,20 +102,28 @@ class Number {
...
     } else {
+      let opts = {
+        country: countryCode,
+        msisdn,
+        target_api_key: targetApiKey,
+        api_key: this.creds.apiKey,
+        api_secret: this.creds.apiSecret
+      };
+
+      if (targetApiKey instanceof Function) {
```
hummusonrails, line 118: I think we should also check here that `callback` is `undefined` and then account for a situation where `targetApiKey` and `callback` are both provided as functions, since this is a relatively new parameter, and that situation is not unlikely.

test/Number-test.js
```diff
@@ -169,6 +169,17 @@ describe("Number", () => {
     );
   });
 
+  it("should allow buying with a target api key", () => {
```
hummusonrails, line 172: Related to the comment above, I'd also add a spec for a failure route when both arguments are defined as functions.

src/Number.js
kellyjandrews: I'll think about how to approach this - the likely-hood of someone providing two functions is really low.

kellyjandrews:
Still a bit more to do on this one, as the `cancel` method also requires the same argument.

src/Number.js
slorello89: I think this is the right way to expand the argument set without breaking the API, or having the callback be anything other than the last arg. This does create a fragrant code smell, so perhaps we could leave a TODO in here to remove this bit of logic for the next major release?

kellyjandrews: This is an unfortunate side effect of callbacks set up in this style.

slorello89 (approved):
looks good to me.

PR #311, opened 2020-07-23
Fix issue with local options being overwritten with Nexmo class

### LINKED ISSUE #295: Nexmo constructor changes of given options object
Here https://github.com/Nexmo/nexmo-node/blob/master/src/index.js#L103
must be copy of an options object instead of direct usage.

kellyjandrews:
Fixes #295

### DIFF
--- a/test/Nexmo-test.js
+++ b/test/Nexmo-test.js
@@ -145,19 +145,24 @@ describe("Nexmo Object instance", function() {
     expect(nexmo.files).to.be.a("object");
   });
 
-  it("should allow options to be passed", function() {
-    var options = {
+  it("should allow options to be passed and remain unchanged", function() {
+    var localOptions = {
       appendToUserAgent: "EXT",
       debug: true
     };
-    new Nexmo(
+
+    let passedOptions = Object.assign({}, localOptions);
+
+    let nexmo = new Nexmo(
       {
         apiKey: "test",
         apiSecret: "test"
       },
-      options
+      passedOptions
     );
-    expect(options.userAgent).to.match(/.*EXT/);
+
+    expect(localOptions).to.eql(passedOptions);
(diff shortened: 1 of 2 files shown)


test/Nexmo-test.js
```diff
@@ -145,19 +145,24 @@ describe("Nexmo Object instance", function() {
...
-  it("should allow options to be passed", function() {
-    var options = {
+  it("should allow options to be passed and remain unchanged", function() {
+    var localOptions = {
       appendToUserAgent: "EXT",
       debug: true
     };
-    new Nexmo(
+
+    let passedOptions = Object.assign({}, localOptions);
```
mheap, line 154: Why do we `Object.assign` here? We should pass `localOptions` to the constructor then assert that they were not changed

kellyjandrews: Because `localOptions === localOptions` regardless, and the test will always pass. I need to compare it to the original without the original changing.

hummusonrails: Can you help me understand, also? From what I am reading this test tests the validity of `Object.assign()` but it's not testing it from within the `constructor` of the Nexmo object. Wouldn't we want our test to test this function call from within the `constructor`, specifically?

hummusonrails: Ah, I see that you are actually testing both in this block. Okay, as long as the `nexmo` object's values are also tested, I'm good with it.

hummusonrails (approved):
I tested this locally, both with the spec and manually and it fixes the issue documented that it is meant to resolve. 🚀

test/Nexmo-test.js
kellyjandrews: @bencgreenberg It's an unfortunate side effect of javascript - you have to store the original as a separate object to validate if the two objects remain the same.

PR #318, opened 2020-08-17
Fix Nexmo is not a constructor error

mheap:
Add add-module-exports Babel module back in and add Travis test to ensure that this works in the future


kellyjandrews (approved):
Changes look good. I'll merge once Travis completes.

PR #678, opened 2022-07-07
Added support for auto-loading known optional modules

dragonmantank:
## Description

This adds the ability for the SDK to load known modules if they are also installed as part of the overall project. The first plugin that will be used this way will be the upcoming Video API. 

## Motivation and Context

Going forward we want to make the beta release much easier to handle, so customers can just install beta packages instead of having one massive beta release. The SDK will then just check to see if any of these optional (but not optional in the NPM sense) packages are installed and register them if they are. If the user is not using an optional package, we just skip over it.

## Testing Details

Bunch of local testing along with the new upcoming packages.

## Types of changes

## Checklist

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,8 +1,7734 @@
 {
   "name": "@vonage/server-sdk",
   "version": "2.12.0-beta.0",
-  "lockfileVersion": 1,
+  "lockfileVersion": 2,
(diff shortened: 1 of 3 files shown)


conshus (approved):
Looks good. Thanks for breaking it down for me and answering my questions.

PR #704, opened 2022-10-04
build: cleanup TS build files

manchuck:
Cleaning up Typescript build files 

## Description
Having multiple `tsconfig.json` files makes it difficult to manage the build process. Having one central config file for all the packages helps reduce DRY config options while still allowing finer control over the build process. There are two different files used to allow separation from building and testing. 

## Motivation and Context
Reduces DRY configs and makes building typescript more manageable 

## Testing Details
Run `tsc --build --verbose` and confirm that all typescript files built without errors

## Types of changes

- [X] Bug fix (non-breaking change which fixes an issue)

## Checklist
- [X] My code follows the code style of this project.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,11 +13,14 @@ jobs:
     runs-on: ${{ matrix.os }}
     steps:
     - uses: actions/checkout@v2
+
     - name: Set up Node.js ${{ matrix.node }}
       uses: actions/setup-node@v1
       with:
         node-version: ${{ matrix.node }}
+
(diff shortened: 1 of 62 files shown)


dragonmantank:
Just noting, while the codecov does fail, not sure why the unit tests fail saying the `Voice` is missing a bunch of dependency classes. I just tried locally and the compile step works fine, but it does die on a unit test for `@vonage/video` (I'm guessing because of a rules change, because this module published fine and we require unit tests to work to publish, and `3.x-wip` still works OK locally at least).

PR #705, opened 2022-10-04
build: fixing actions

manchuck:
Fixing Github actions

## Description
Just adding in missing actions and increasing performance

## Motivation and Context
Not having actions running prevents other merges from happening

## Testing Details
If the actions pass for this PR, then it works

## Types of changes

- [X] Bug fix (non-breaking change which fixes an issue)

## Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,11 +1,8 @@
-# This workflow will do a clean install of node dependencies, build the source code and run tests across different versions of node
-# For more information see: https://help.github.com/actions/language-and-framework-guides/using-nodejs-with-github-actions
-
 name: Vonage
 
 on:
   push:
-    branches: [ master ]
+    branches: [ 3.x-wip ]
   pull_request:
 jobs:
   test:
@@ -23,4 +20,4 @@ jobs:
     - name: Install dependencies
       run: npx lerna bootstrap
     - name: Lint, Compile, Test
-      run: npx lerna run compile && npx lerna run test
+      run: npx lerna run compile && npm run test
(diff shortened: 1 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -1,11 +1,8 @@
-# This workflow will do a clean install of node dependencies, build the source code and run tests across different versions of node
-# For more information see: https://help.github.com/actions/language-and-framework-guides/using-nodejs-with-github-actions
-
 name: Vonage
 
 on:
   push:
-    branches: [ master ]
+    branches: [ main ]
```
dragonmantank: This should be `3.x-wip` for the moment, this repo doesn't have a `main` branch. Long story, but the "main" branch is currently `2.x`

.github/workflows/codecov.yml
```diff
@@ -5,14 +5,14 @@ name: Codecov Report
 
 on:
   push:
-    branches: [ master ]
+    branches: [ main ]
```
dragonmantank: Same as above, this should target `3.x-wip`

package.json
```diff
@@ -27,10 +27,12 @@
   ],
   "scripts": {
     "bootstrap": "npx lerna bootstrap",
-    "lerna": "npx lerna",
-    "publish": "npx lerna publish",
-    "test": "npx lerna run test",
-    "test-watch": "./node_modules/.bin/jest --watch"
+    "lerna": "./node_modules/.bin/lerna",
```
dragonmantank, line 30: Is there an advantage to hard-coding this versus using `npx`?

dragonmantank (changes requested):

package.json
manchuck: Yea, since Lerna is installed as a dev dependency, then NPX will not have to download it. This makes the actions faster as it doesn't have to download to node_modules with npm and then again with NPX. 

Come to think of it; we should cache `node_modules` for faster execution as well.

Also, the `test` script is going to be updated to use jest instead of Lerna in another branch

PR #721, opened 2022-11-10
docs: removed duplicate docs 

manchuck:
Removed Make file and duplication COC, Contributing and License files. It helps keep the root cleaner and prevents the duplication of these files


.github/CONTRIBUTING.md
```diff
@@ -27,3 +27,15 @@ To run the code for an open PR, follow these steps:
 1. Run the sample code snippet to test the SDK.
 
 When reviewing PRs, and switching branches to compare running examples against feature branches, don't forget to run `npm compile` after switching branches.
+
+## Maintainers
+
+The following is a quick cheatsheet for maintainers.
+
+* New commits are pushed to their own feature branch and merged into `master`.
```
dragonmantank: Need to rework this, as this isn't how we actually release, and the branch names need corrected.

dragonmantank (changes requested):
Need to clean up the maintenance stuff

Not sure we should move these files to `.github/` ... convention is to generally keep these things in the root of the project, and leave `.github/` to files that are specific only to GitHub functionality. 

While not a blocker, I would prefer to keep a `MAINTENANCE.md` file separate from contributing, as contributors only need to worry about how to submit PRs.

.github/CONTRIBUTING.md
manchuck: Ok let us chat about this a bit more when we want to make the 3.x the main branch

manchuck:
@dragonmantank moving everything to `.github` makes packaging easer, as we can ignore everything in that directory. Another reason is that we had some of the docs copied into the packages which could cause confusion

dragonmantank:
I'd prefer we keep the main docs in the root, as that is where people expect them. I don't mind removing the duplicates from the sub-packages as we can link to them the READMEs, but we shouldn't push the files off to somewhere else. None of the root stuff is pulled into the packages anyway.

dragonmantank (approved):

PR #728, opened 2022-11-10
build: eslint config cleanup

manchuck:
Consolidated eslint config files and removed tslint

PR #729, opened 2022-11-11
build: updated dependencies

manchuck:
Removed unused dependencies and fixed the scripts for the child packages


manchuck:
requires #727

dragonmantank (changes requested):
Was this based on an older branch? All the packages split out `auth`, `server-client`, and `vetch`, but those should be all wrapped up in a single dependency of `server-client`. A lot of them are also updating from `server-client:1.0.0` to `1.0.1`, which already happened in the main 3.x branch.

manchuck:
@dragonmantank This was based on the `3.x` branch. I used a tool called [depcheck](https://www.npmjs.com/package/depcheck) on each of the packages to ensure that we have the correct dependencies when installing the packages on their own (`lerna exec --no-bail --prefix --stream npx depcheck`). It checks the import/require statements to ensure that the package.json is not missing anything.

manchuck:
This technically requires both #727 and #728 as those clean up the linter and typescript configs. Although this should not impact development on its own

PR #733, opened 2022-11-17
style: linter fixes

manchuck:


manchuck:
needs #729

manchuck:
No longer going to have one big branch that addresses all the linter issues. Instead, we should clean up each package on its own to avoid a crazy long PR (we have been letting the checks fail while we get the `3.x` branch working anyways)

Issue #742, opened 2022-11-24
Some typescript types are wrong


maximelebastard:
## Expected Behavior
I follow the documentation and the types to implement Vonage SMS sendings. I'd like to rely on them.
I have two basic examples that don't work:

A/ The Authentication

The getting started on the console is the following to send a basic SMS

<img width="975" alt="Capture d’écran 2022-11-24 à 15 36 42" src="https://user-images.githubusercontent.com/6564357/203812500-dd411070-f1bd-41f5-932b-3a017f1fb968.png">

Therefore, I expect the constructor type to look like this - with only apiKey and apiSecret to be required

```
export declare class Vonage {
    constructor(credentials: AuthInterface);
}

export interface AuthInterface {
    apiKey: string;
    apiSecret: string;
    signature?: SignedHashParams;
    applicationId?: string;
    privateKey?: string;
    getQueryParams<T>(params?: T)?: AuthQueryParams & T;
    createSignatureHash<T>(params: T)?: AuthSignedParams & T;
    createBasicHeader()?: string;
    createBearerHeader()?: string;
}
```

B/ The response of the sending

I'd like to get a sending ID after I've sent the SMS, and I saw this type for the sms.send response
```
export interface SendSMSResponse {
    'message-count': string;
    messages: Message[];
}
export interface Message {
    to?: string;
    messageId?: string;
    status?: string;
    remainingBalance?: string;
    messagePrice?: string;
    network?: string;
    clientRef?: string;
    accountRef?: string;
}
```

Therefore, I expect to be able to use `res.messages[0].messageId` to get my the message ID I need

## Current Behavior

A/ The Authentication

If I bascially try your example, I get the following error

```
Argument of type '{ apiKey: string; apiSecret: string; }' is not assignable to parameter of type 'AuthInterface'.
  Type '{ apiKey: string; apiSecret: string; }' is missing the following properties from type 'AuthInterface': getQueryParams, createSignatureHash, createBasicHeader, createBearerHeaderts(2345)
```

This is because all the methods are indeed required in your definition.

B/ The response of the sending

The case of your type is wrong
```
messages?.[0].messageId // = undefined
messages?.[0]["message-id"] // = my message ID
```

Fortunately, I had some unit tests to spot that and it didn't go to production like that

## Potential solutions

I will probably check every of the types you provide, and put some `any` casts anywhere they're not right.
But these are very basic examples, it makes me consider your typescript definitions cannot be trusted - which is very annoying.

[...]

PR #743, opened 2022-11-26
fix: Correct our return types when sending SMS

dragonmantank:
## Description

Corrects the type definition for the returned SMS responses to match the original declaration. This also adds keys with appropriate deprecation notices to not break existing code. 

## Motivation and Context

Per #742 , the definitions we provided for the return response did not match the API. Looks like as this was an early module, we never finished cleaning up to make it match.

## Testing Details

Unit tests

## Types of changes

## Checklist

### DIFF
--- a/packages/sms/__tests__/sms.test.ts
+++ b/packages/sms/__tests__/sms.test.ts
@@ -60,5 +60,44 @@ describe('SMS', () => {
             text: 'Sample SMS Text',
         })
         expect(results['message-count']).toEqual('1')
+        expect(results.messageCount).toEqual(1)
+        for (let i = 0; i < results.messages.length; i++) {
+            expect(resp.messages[i].to).toEqual(results.messages[i].to)
+            expect(resp.messages[i]['message-id']).toEqual(
+                results.messages[i].messageId
+            )
+            expect(resp.messages[i]['message-id']).toEqual(
+                results.messages[i]['message-id']
+            )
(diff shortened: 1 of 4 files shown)


packages/sms/__tests__/sms.test.ts
```diff
@@ -60,5 +60,44 @@ describe('SMS', () => {
...
+                results.messages[i].clientRef
+            )
+            expect(resp.messages[i]['account-ref']).toEqual(
+                results.messages[i].accountRef
+            )
+            expect(resp.messages[i]['account-ref']).toEqual(
+                results.messages[i]['account-ref']
+            )
+        }
+        expect(results.messageCount).toEqual(1)
```
manchuck, line 101: This test should be done using `test.each`. Otherwise, we call the `expects` conditionally if `results.messages.length` is empty. This could happen with a bad mock

manchuck (changes requested):
Just update the test

packages/sms/__tests__/sms.test.ts
dragonmantank: You'd never have a `results.messages.length` < 1, because it's a list of the messages that were sent. A length of 0 would indicate we somehow sent nothing, but were still successful. The above test isn't testing different use cases, all the `expects()` are a single valid response being returned, and testing each property (mostly for backward compatibility)

manchuck (approved):
Let's go with it for now, and when we add in the linter rules, we can most likely make it an exception if it bugs out

PR #787, opened 2023-01-12
feat: audit package

manchuck:
Adding the Audit module

## Description
Creating a new package for interacting with the audit API. I also added some transform functions for objects as I noticed the same functionality was duplicated in other places and won't go deep into the object. These functions have been added to `server-client` along with a new type that defines our API links 

## Types of changes

## Checklist


packages/audit/lib/enums.ts
```diff
@@ -0,0 +1,29 @@
+export enum AuditEventTypes {
```
dragonmantank, line 1: To keep with SDK convention, I'd break this out to `enums/AuditEventTypes.ts`. It's just the older modules that still use one combined file.

packages/audit/lib/types.ts
```diff
@@ -0,0 +1,42 @@
+import { VetchResponse, VetchOptions } from '@vonage/vetch';
```
dragonmantank: To keep with SDK convention, I'd break this out to individual types in a `types/` folder. It's just the older modules that still use one combined file.

packages/server-sdk/lib/vonage.ts
```diff
@@ -1,56 +1,58 @@
...
-    public numberInsights: NumberInsights
-    public numbers: Numbers
-    public pricing: Pricing
-    public secrets: Secrets
-    public sms: SMS
-    public verify: Verify
-    public voice: Voice
+  public accounts: Accounts;
+  public applications: Applications;
+  public audit: Audit;
```
dragonmantank, line 28: Since Audit is considered a beta product, the user will need to instantiate it manually. Beta products will always be "opt-in" so we can avoid having a beta SDK and get into the mess we had with the original beta branches.

packages/server-sdk/package.json
```diff
@@ -29,6 +29,7 @@
     "dependencies": {
         "@vonage/accounts": "^1.0.13",
         "@vonage/applications": "^1.0.12",
+        "@vonage/audit": "^1.0.0",
```
dragonmantank, line 32: Would need to remove this - the developer will opt-in in their own app to use Audit/any beta products

dragonmantank (changes requested):
Breaking some things out, and removing Audit as an actual dependency

dragonmantank (approved):
LGTM

PR #794, opened 2023-02-21
fix(types): verify response types

manchuck:
This fixes the typescript interfaces and types for the verify module

## Description

I noticed some other issues with some of the types being returned as well as the requests being passed into the verify module. I have addressed them all and updated the tests 

## Types of changes

## Checklist


packages/verify/lib/classes/PSD2.ts
```diff
@@ -1,52 +1,54 @@
...
-    nextEventWait?: number
-    workflowId?: number
+export class PSD2 implements PSD2Parameters {
+  number: string;
+  payee: string;
+  amount: number;
+  country?: string;
+  codeLength?: number;
+  lg?: VerifyLanguages;
+  language?: string;
```
dragonmantank, line 14: We should mark this as deprecated, and will still need to take this into account when serializing down to a JSON blob

packages/verify/lib/classes/PSD2.ts
```diff
@@ -1,52 +1,54 @@
...
+      this.payee = payee;
+    }
+    if (amount) {
+      this.amount = amount;
+    }
+    if (codeLength) {
+      this.codeLength = codeLength;
+    }
+    if (language) {
+      this.lg = language;
```
dragonmantank, line 46: We should still set `this.language` for the time being as well

packages/verify/lib/classes/Verification.ts
```diff
@@ -1,50 +1,51 @@
...
-    pinExpiry?: number
-    nextEventWait?: number
-    workflowId?: number
+export class Verification implements VerificationParameters {
+  number: string;
+  brand: string;
+  country?: string;
+  senderId?: string;
+  codeLength?: number;
+  lg?: VerifyLanguages;
```
dragonmantank, line 14: We should keep `language` for BC and mark it as deprecated

packages/verify/lib/classes/Verification.ts
```diff
@@ -1,50 +1,51 @@
...
+      this.country = country;
     }
+    if (senderId) {
+      this.senderId = senderId;
+    }
+    if (codeLength) {
+      this.codeLength = codeLength;
+    }
+    if (language) {
+      this.lg = language;
```
dragonmantank, line 44: Same as PSD2, should also set `this.language` for the time being

packages/verify/lib/enums/Workflows.ts
```diff
@@ -0,0 +1,9 @@
+export enum VerifyWorkflows {
+    SMS_TTS_TTS = '1',
+    SMS_SMS_TTS = '2',
+    TTS_TTS = '3',
+    SMS_SMS = '4',
+    SMS_TTS = '5',
+    SMS = '6',
+    TMS = '7',
```
dragonmantank: This should be `TTS`

packages/verify/lib/verify.ts
```diff
@@ -1,120 +1,91 @@
...
       cmd: Command.CANCEL,
     };
 
-    const resp = await this.sendPostRequest<VerifyControlResponse>(
-      `${this.config.apiHost}/verify/control/json`,
-      data,
-    );
-    return resp.data;
+    // eslint-disable-next-line max-len
+    const resp = await this.sendPostRequest<
```
dragonmantank, line 33: This is probably going to be in a few different places, but for consistency I think we should actually throw an Error instead of returning an error. This is more consistent with the newer APIs work, and much easier to find and handle than returning an error.

That would be a BC break, so I would document that behavior change, or make it forward compatible with an options flag that will allow you to keep existing behavior or throw an error (thereby deprecating the old behavior and allowing the user a way to opt-in to new behavior to make upgrading easier). 

Just because the API is bad and doesn't return proper HTTP codes doesn't mean the SDK shouldn't behave as expected.

dragonmantank (changes requested):

packages/verify/lib/verify.ts
manchuck: I want to avoid having this PR being a BC one. This was trying to address issues with the types. When we start working on verify V2, then we'll go back and have this throw an exception.

dragonmantank (approved):
LGTM

PR #795, opened 2023-02-23
feat: adding debug logging

manchuck:
Adding debug logging 

## Types of changes

## Checklist

### DIFF
--- a/jest.config.ts
+++ b/jest.config.ts
@@ -15,10 +15,6 @@ const config: Config.InitialOptions = {
     },
     {
       displayName: 'AUDIT',
-      testMatch: ['<rootDir>/packages/audit/__tests__/**/*.test.ts'],
-    },
-    {
-      displayName: 'AUDIT',
       testMatch: ['<rootDir>/packages/audit/tests__/**/*.test.ts'],
     },
     {
--- a/jest.config.ts
+++ b/jest.config.ts
@@ -15,7 +15,7 @@ const config: Config.InitialOptions = {
     },
     {
       displayName: 'AUDIT',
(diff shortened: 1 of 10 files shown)


dragonmantank (approved):
LGTM

jest.config.ts
```diff
@@ -13,10 +13,6 @@ const config: Config.InitialOptions = {
         '<rootDir>/packages/applications/__tests__/**/*.test.ts',
       ],
     },
-    {
```
dragonmantank: Did you mean to remove the Audit tests?

packages/vetch/lib/vetch.ts
```diff
@@ -60,7 +61,8 @@ export class Vetch {
       let data = await res.text();
       try {
         data = JSON.parse(data);
-      } catch {
+      } catch (error) {
+        log('Invalid JSON from API', error);
```
dragonmantank: This isn't necessarily an error that we need to log, because not all the APIs will return JSON 100% of the time. Media and Reports API are two that will regularly not return JSON.

Basically we just want to parse JSON if it's JSON, but otherwlse leave it alone. Doing it in a try/catch was just an easy way to parse JSON without having to make sure the API returns a proper header.

dragonmantank (changes requested):

jest.config.ts
dragonmantank: After looking at this again, I think we should remove the one below and put this back (or fix the one below). The path to the audit tests is wrong in the one that is left

dragonmantank (changes requested):
One more small tweak

dragonmantank (approved):
LGTM once the CI finishes

PR #798, opened 2023-02-27
feat: redact module

manchuck:
Provides SDK access for the redact API 

## Types of changes

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,25 +39,25 @@ information for each package below:
 
 The following is a list of Vonage APIs and whether the Node Server SDK provides support for them:
 
-| API   | API Release Status |  Supported?|
-|----------|:---------:|:-------------:|
-| Account API | General Availability |✅|
-| Alerts API | General Availability |✅|
-| Application API | General Availability |✅|
-| Audit API | Beta |✅|
-| Conversation API | Beta |❌|
-| Dispatch API | Beta |❌|
-| External Accounts API | Beta |❌|
-| Media API | Beta |❌|
-| Messages API | Beta |✅|
-| Number Insight API | General Availability |✅|
-| Number Management API | General Availability |✅|
-| Pricing API | General Availability |✅|
-| Redact API | Developer Preview |❌|
-| Reports API | Beta |❌|
-| SMS API | General Availability |✅|
-| Verify API | General Availability |✅|
-| Voice API | General Availability |✅|
+| API                   |  API Release Status  | Supported? |
+|-----------------------|:--------------------:|:----------:|
+| Account API           | General Availability |     ✅     |
+| Alerts API            | General Availability |     ✅     |
+| Application API       | General Availability |     ✅     |
+| Audit API             |         Beta         |     ✅     |
+| Conversation API      |         Beta         |     ❌     |
+| Dispatch API          |         Beta         |     ❌     |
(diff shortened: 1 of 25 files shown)


README.md
```diff
@@ -53,7 +53,7 @@ The following is a list of Vonage APIs and whether the Node Server SDK provides
 | Number Insight API | General Availability |✅|
 | Number Management API | General Availability |✅|
 | Pricing API | General Availability |✅|
-| Redact API | Developer Preview |❌|
+| Redact API | General Availability |✅|
```
dragonmantank: Redact is still Developer Preview, but we can mark it that we now support it

packages/redact/lib/interfaces/response/errorResponse.ts
```diff
@@ -0,0 +1,6 @@
+export interface ErrorResponse {
```
dragonmantank, line 1: Wouldn't hold this up, but since we try and return RFC 7807 responses like this, we could move to the vetch package maybe, so we don't have to do this every API.

Alternate argument - we don't define this and let it trickle up via Error handling instead

packages/server-sdk/README.md
```diff
@@ -117,10 +117,12 @@ The following is a list of Vonage APIs and whether the Node Server SDK provides
 | Number Insight API | General Availability |✅|
 | Number Management API | General Availability |✅|
 | Pricing API | General Availability |✅|
-| Redact API | Developer Preview |❌|
+| Redact API | General Availability |✅|
```
dragonmantank: Same as before, keep this "Developer Preview"

We're only making it available via the kitchen sink because we had it in 2.x

packages/server-sdk/lib/vonage.ts
```diff
@@ -44,12 +46,13 @@ export class Vonage {
     this.options = options;
 
     this.accounts = new Accounts(this.credentials, this.options);
-    this.audit = new Audit(this.credentials, this.options);
     this.applications = new Applications(this.credentials, this.options);
+    this.audit = new Audit(this.credentials, this.options);
```
dragonmantank: The Audit module shouldn't be brought in as it's in Beta status right now

dragonmantank (changes requested):
Pull out audit, and some minor text changes

packages/redact/lib/interfaces/response/errorResponse.ts
manchuck: Yea I agree. Next time we give vetch a little TLC, we can visit

packages/server-sdk/package.json
```diff
@@ -27,18 +27,18 @@
         "compile": "npx tsc --build --verbose"
     },
     "dependencies": {
-        "@vonage/accounts": "^1.1.2",
```
dragonmantank, line 30: Do we need to rebase here? We've ended up dropping back on all these packages.

If this is going to get cleaned up before deploy we can ignore this for now.

dragonmantank (approved):
Provisionally approving depending on if we need to fix the versions in the server-sdk `package.json`

packages/server-sdk/package.json
manchuck: Yea, once this and the SMS PR gets merged in, I will clean up the packages.

PR #801, opened 2023-03-02
feat(auth,messages,server-client): Made auth handlers async

dragonmantank:
## Description

Allow auth handlers to be overridden in case custom handlers need to be used

## Motivation and Context

The video team needs an alternate way to get a JWT token, so this allows them to supply a custom auth handler. Expanded this to also work with the other non-JWT handlers as well to be consistent.

Messages also gets touched because it was the one case we overrode how to add auth to a request.

## Testing Details

Unit tests

## Types of changes

## Checklist

### DIFF
--- a/packages/auth/lib/types.ts
+++ b/packages/auth/lib/types.ts
@@ -38,8 +38,8 @@ export interface AuthInterface {
     signature?: SignedHashParams
     applicationId?: string
     privateKey?: string
-    getQueryParams<T>(params?: T): AuthQueryParams & T
-    createSignatureHash<T>(params: T): AuthSignedParams & T
-    createBasicHeader(): string
-    createBearerHeader(): string
+    getQueryParams<T>(params?: T): Promise<AuthQueryParams & T>
+    createSignatureHash<T>(params: T): Promise<AuthSignedParams & T>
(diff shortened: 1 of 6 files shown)


packages/auth/lib/types.ts
```diff
@@ -38,8 +38,8 @@ export interface AuthInterface {
...
     applicationId?: string
     privateKey?: string
-    getQueryParams<T>(params?: T): AuthQueryParams & T
-    createSignatureHash<T>(params: T): AuthSignedParams & T
-    createBasicHeader(): string
-    createBearerHeader(): string
+    getQueryParams<T>(params?: T): Promise<AuthQueryParams & T>
+    createSignatureHash<T>(params: T): Promise<AuthSignedParams & T>
+    createBasicHeader(): Promise<string>
+    createBearerHeader(): Promise<string>
```
manchuck, line 44: This might be considered a BC lets keep that in mind

dragonmantank: I thought about that, and I think it's OK to do as a point release. These are mostly internal, and no customer should have ever had a reason to override any of those methods. We should point it out in the release notes, but I think we'll be safe.

manchuck (approved):

PR #805, opened 2023-03-15
feat: verify v2

manchuck:
Adding in Verify v2

## Description
This adds in the Verify V2 module 

## Types of changes

## Checklist


packages/verify2/lib/verify2.ts
```diff
@@ -0,0 +1,34 @@
+import { AuthenticationType, Client } from '@vonage/server-client';
+import {
+  Request,
+  VerificationRequestParams,
+  VerificationResponse,
+} from './types';
+
+export class Verify2 extends Client {
+  protected authType = AuthenticationType.JWT;
```
dragonmantank, line 9: Verify v2 also supports basic auth headers, it just disables the callbacks

dragonmantank (changes requested):
Verify v2 also supports basic header auth, so like with Messages we should support that (it's not a side effect of the auth changes, but a way for customers to still send Verify without setting up an application)

packages/verify2/lib/verify2.ts
manchuck: Resolved offline

dragonmantank (approved):

PR #818, opened 2023-04-24
feat: proactive connect

manchuck:
Adding in proactive connect

## Types of changes

## Checklist


packages/proactive-connect/README.md
```diff
@@ -0,0 +1,84 @@
...
+
+const credentials = new Auth({
+  applicationId: APP_ID,
+  privateKey: PRIVATE_KEY_PATH,
+});
+const options = {};
+const proactiveConnectClient = new ProactiveConnect(credentials, options);
+```
+
+Where `credentials` is any option from [`@vonage/auth`](https://github.com/Vonage/vonage-node-sdk/blob/3.x/packages/auth/README.md#options), and `options` is any option from [`@vonage/server-client`](https://github.com/Vonage/vonage-node-sdk/blob/3.x/packages/server-client/README.md#options)
```
dragonmantank, line 47: Does this work with all credential types or just App/Key?

packages/proactive-connect/lib/proactiveConnect.ts
```diff
@@ -0,0 +1,194 @@
...
+    writeFileSync(file, resp.data);
+    return true;
+  }
+
+  async importListItems(
+    listId: string,
+    file: string,
+  ): Promise<ImportFileResponse> {
+    const itemsForm = new FormData();
+    itemsForm.setBoundary(this.FORM_BOUNDARY);
```
dragonmantank, line 183: Is there not a library that can handle creating multipart form bodies?

dragonmantank (changes requested):
Just a couple of questions, otherwise looks good

packages/proactive-connect/README.md
manchuck: Since we are using `server-client`, it is best to just keep the documentation the same

packages/proactive-connect/lib/proactiveConnect.ts
manchuck: This lib does, but it creates a different form on each instance. We set the form boundary here for testing purposes. I went through several libs to parse multi-parts, and they were all crap except for the one we are using. However that one needs the boundary explicitly set.

PR #821, opened 2023-04-28
refactor: added custom exceptions when required parameters are missing

manchuck:
## Types of changes

## Checklist


packages/auth/lib/auth.ts
```diff
@@ -28,44 +34,86 @@ export class Auth implements AuthInterface {
...
+      this.privateKey,
+      this.jwtOptions,
+    )}`;
   };
 
   createSignatureHash = async <T>(
     params: AuthSignedParams & T,
   ): Promise<AuthSignedParams & T> => {
     log('Creating signature hash');
+    if (!this.apiKey) {
```
dragonmantank, line 111: This method should also make sure they supplied a signature hash. If there is no signature string then we can't generate a signature.

dragonmantank (changes requested):
The new changes look OK except I think we should be checking for API key and signature string in the signature method, not just the signature stuff.

This also seems to be built off the Proactive Connect code, can that be stripped out since this PR has nothing to do with PC?

manchuck:
@dragonmantank That was a strange rebase that happened. Either way, some GitFu and PC have been removed

dragonmantank (approved):
LGTM

PR #835, opened 2023-06-07
feat(voice): adding advanced machine detection

manchuck:
## Types of changes

## Checklist


packages/voice/__tests__/ncco.lagecy.test.ts
```diff
@@ -0,0 +1,68 @@
+import { NCCOBuilder } from '../lib';
```
dragonmantank: File is typo'd, should be "legacy"

Not sure why this is called "legacy" though, this is testing the NCCO builder, which is valid?

dragonmantank (approved):

packages/voice/__tests__/ncco.lagecy.test.ts
manchuck: oh yea now that we are keeping these tests it no longer needs to be this. I was keeping it to make sure I wasn't introducing breaking changes

PR #838, opened 2023-06-22
refactor(voice): cleanup voice package

manchuck:
## Types of changes

## Checklist


packages/voice/lib/classes/NCCO/Connect.ts
```diff
@@ -1,99 +1,109 @@
...
+import { NCCOActions } from '../../enums';
+import { ConnectEventType } from '../../enums/NCCO/ConnectEventType';
+import { MachineDetection } from '../../enums/NCCO/MachineDetection';
+import { ConnectAction } from '../../types/NCCO/ConnectAction';
+import { Serializable } from '../../ncco';
+import { CallEndpoint } from '../../types/Endpoint/CallEndpoint';
+import debug from 'debug';
 
+debug('@vonage/voice')(
+  'This class is deprecated. Please update to use the appropriate type',
```
dragonmantank: It's inappropriate to deprecate these and replace with a type. The point of the classes is to make it easier to make NCCO objects, handling things like arrays, setting the type, etc, hence the `serializeToNCCO()` method.

You could have it probably alias a type, but the point of these are to be actual classes (unlike the argument where you could swap many interfaces around for types).

dragonmantank (changes requested):

dragonmantank (approved):

PR #845, opened 2023-07-26
feat: users package

manchuck:
## Types of changes

## Checklist


packages/user/lib/user.ts
```diff
@@ -0,0 +1,143 @@
...
+
+  async getUser(userId: string): Promise<UserType> {
+    const resp = await this.sendGetRequest<UserResponse>(
+      `${this.config.apiHost}/v1/users/${userId}`,
+    );
+
+    return apiUserToUser(resp.data);
+  }
+
+  async updateUser(userId: string, user: UserType): Promise<UserType> {
```
dragonmantank: This should just take a UserType, and bail out if they don't pass an ID as part of the object. There's never a reason to update a user without an ID, and we don't want someone passing in an ID trying to update a user object with a different ID
```suggestion
  async updateUser(user: UserType): Promise<UserType> {
```

packages/user/lib/user.ts
```diff
@@ -0,0 +1,143 @@
...
+    const resp = await this.sendPutRequest<UserResponse>(
+      `${this.config.apiHost}/v1/users/${userId}`,
```
dragonmantank:
```suggestion
      `${this.config.apiHost}/v1/users/${user.id}`,
```

dragonmantank (changes requested):
Looks good, would change the update user method though

packages/user/lib/user.ts
manchuck: Ive been wanting to have the methods go this way but didn't because that was how the other packages had the signature this way

dragonmantank (approved):

PR #850, opened 2023-08-08
fix: timeout parameter being respected

### LINKED ISSUE #843: Request timeout option is broken
I'm trying to interrupt long request by leveraging the timeout option available on the Vonage client object but it does not behave as expected.

## Expected Behavior
The request should throw a timeout exception when said request is taking too much time.

## Current Behavior
No exception is raised and the request keeps running beyond the timeout limitation.

## Possible Solution
The timeout option is explicitly ignored in the code as you can see [here](https://github.com/Vonage/vonage-node-sdk/blob/350ad18452d4e6bf2936a76a8b462c6018241eb3/packages/server-client/lib/client.ts#L45)

## Steps to Reproduce (for bugs)
Here is a code snippet describing the usage and the expected behavior
```typescript
import { Auth } from '@vonage/auth';
import { Vonage } from '@vonage/server-sdk';
import type { SMSMessages } from '@vonage/sms/dist/types';

const requestTimeout = 1; // in milliseconds I supposed. It is nowhere to be found in documentation
const client: Vonage = new Vonage(
  new Auth({
    apiKey: <my-api-key>,
    apiSecret: <my-api-secret>,
  }),
  {
    responseType: undefined,
    restHost: <vonage-api-host>,
    timeout: requestTimeout, // This timeout option is being completely ignored
    videoHost: undefined,
    apiHost: undefined,
  },
);

const sendSMS = (phone: string, message: string, from: string): Promise<SMSMessages> =>
{
  return client.sms.send({
    to: phone,
    from,
    text: message,
    clientRef: config.clientName,
  });
}

// This following call should fail due to timeout parameter being provided with a small value
// But it always succeeds, completely ignoring the timeout option provided to the client
sendSMS('1233456987', 'Message that shouldn't get sent because of timeout', 'Timeout')
  .then((res) => console.log(res))
  .catch((error) => console.log(error));
``` 

## Context
We are trying to avoid SMS sending jobs locking our dedicated fast running worker due to requests taking too much time to be executed.

## Your Environment

* SDK Version: 3.2.0
* Node Version: 18.15.0
* Operating System and version: alpine:3

manchuck:
resolves #843 

## Types of changes

## Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,35 +13,21 @@ jobs:
       matrix:
         os: [ubuntu-latest, windows-latest, macos-latest]
         node: [16.x, 18.x, 20.x]
-        experimental: [false]
-        include:
-          - os: ubuntu-latest
-            node: 14
-            experimental: true
-            npm_version: "@8.3.1"
-          - os: windows-latest
-            node: 14
-            experimental: true
-            npm_version: "@8.3.1"
-          - os: macos-latest
-            node: 14
-            experimental: true
-            npm_version: "@8.3.1"
     steps:
-    - uses: actions/checkout@v3
+      - uses: actions/checkout@v3
 
-    - name: Set up Node.js ${{ matrix.node }}
-      uses: actions/setup-node@v3
-      with:
-        node-version: ${{ matrix.node }}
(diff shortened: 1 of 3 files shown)


dragonmantank (approved):
LGTM

PR #857, opened 2023-08-16
feat(media): support for the media api

manchuck:
## Types of changes

## Checklist


packages/server-client/lib/client.ts
```diff
@@ -109,10 +105,11 @@ export abstract class Client {
   public async sendFormSubmitRequest<T>(
     url: string,
     payload?: { [key: string]: any },
+    method?: HTTPMethods.POST | HTTPMethods.PUT,
```
dragonmantank: Are we sure this is correct? The [spec](https://developer.vonage.com/en/api/media#update-a-media-item) mentions that it's a PUT request, but then the example it shows is a POST request, and generally you would do a PUT with a form-encoded body.

dragonmantank (changes requested):

dragonmantank (approved):
LGTM

PR #858, opened 2023-08-16
feat: file download

manchuck:
## Types of changes

## Checklist


packages/server-client/README.md
```diff
@@ -2,16 +2,18 @@
...
+This is the Vonage Server Client SDK for Node.js used to wrap the authentication
+headers/signatures for use with [Vonage APIs](https://www.vonage.com/). To use
+it you will need a Vonage account. Sign up [for free at vonage.com][signup].
 
 For full API documentation refer to [developer.vonage.com](https://developer.vonage.com/).
 
 * [Installation](#installation)
 * [Usage](#usage)
-    * [Options](#options)
+  * [Options](#options)
```
dragonmantank, line 15: Spacing issue? or just a weird GH visual thing?

dragonmantank (changes requested):
We should also add this to voice with something like `vonage.voice.downloadFile()` as a wrapper, since that's almost all its used for. That makes it easier for the developer to use as they won't have to instantiate their own client directly, it can use their voice credentials.

packages/server-client/README.md
manchuck: I must have updated this when the `.editorconfig` has tabs set to 4 instead of 2. It still looks fine in GH

dragonmantank (approved):

PR #875, opened 2023-10-18
docs: adding server-client doc blocks

manchuck:
This also updated vetch as it was not providing the functionality needed. Vetch is now deprecated and only contains type definitions for server-client 

## Types of changes

## Checklist


packages/server-client/lib/enums/AuthenticationType.ts
```diff
@@ -1,7 +1,33 @@
...
     BASIC = 'basic',
+
+    /**
+     * @description JSON Web Token (JWT) authentication.
+     */
     JWT = 'jwt',
+
+    /**
+     * @description Authentication using both API key and secret in the request body.
+     * @deprecated This method is deprecated.
```
dragonmantank, line 20: This is only deprecated in certain APIs, not across-the-board

packages/server-client/lib/enums/GenericErrorCodes.ts
```diff
@@ -1,21 +1,139 @@
...
+  /**
+   * The request originates from an unauthorized IP address. The source IP
+   * address of the request is not in our allow list. Make a request from an
+   * allowed IP address, or add your current IP to the list of authorized
+   * addresses.
+   */
+  INVALID_IP = 'invalid-ip',
+
+  /**
+   * Multiple authentication methods were provided in the request. Multiple
```
dragonmantank: duplicated lines basically

packages/server-client/lib/fileClient.ts
```diff
@@ -1,30 +1,90 @@
...
 const log = debug('vonage:server-client');
 
+/**
+ * A client for downloading files from Vonage.
+ *
+ * @extends Client
+ */
 export class FileClient extends Client {
-  protected authType = AuthenticationType.JWT;
+  public authType = AuthenticationType.JWT;
```
dragonmantank, line 17: Should we be letting this change directly, or add a setter?

packages/server-client/lib/fileClient.ts
```diff
@@ -1,30 +1,90 @@
...
 
+  protected saveFilePath: string = '';
```
dragonmantank, line 19: Is this async-safe? For example, if I'm downloading a batch of different files

packages/server-client/lib/client.ts
```diff
@@ -14,17 +15,50 @@ import { ConfigParams } from './types/index';
...
 
-  protected authType?: AuthenticationType;
+  /**
+   * The type of authentication used for the client's requests.
+   *
+   * @type {AuthenticationType | undefined}
+   * @default AuthenticationType.QUERY_KEY_SECRET
+   * @public
+   */
+  public authType?: AuthenticationType = AuthenticationType.QUERY_KEY_SECRET;
```
dragonmantank, line 34: Not sure why we are switching these to public, the user should never be changing the authentication type by themselves

packages/server-client/lib/client.ts
```diff
@@ -39,15 +73,28 @@ export abstract class Client {
       meetingsHost: options?.meetingsHost || 'https://api-eu.vonage.com',
       proactiveHost: options?.proactiveHost || 'https://api-eu.vonage.com',
       responseType: options?.responseType || ResponseTypes.json,
-      timeout: options?.timeout || null,
+      timeout: options?.timeout || 3000,
```
dragonmantank: Not sure why we are setting this now - we should only set a value if the user explicitly adds a timeout. Otherwise, we may cut connections early for anyone taking advantage of the default behavior

packages/server-client/lib/client.ts
```diff
@@ -60,41 +107,52 @@ export abstract class Client {
...
       });
       return request;
+    }
 
-    case AuthenticationType.QUERY_KEY_SECRET:
-      requestPath = 'params';
-      // falls through
-    case AuthenticationType.KEY_SECRET:
-    default:
+    // Regardless of the setting, set the auth params in the query string
```
dragonmantank: Why this change? We shouldn't be trying to add a key/secret to the query params based on request type, we should be adding them based on the API's expected authentication. 

Also HTTP/1.1 _does_ allow a `GET` and `DELETE` request with a body, so we can't use that as a determining factor in adding to the query params. You shouldn't, but you can (and I don't think any of our APIs do it, but it's technically within spec)

packages/server-client/lib/client.ts
```diff
@@ -60,41 +107,52 @@ export abstract class Client {
...
-      !request[requestPath]
-      && this.authType !== AuthenticationType.QUERY_KEY_SECRET
-    ) {
-      requestPath = 'body';
-      params = new URLSearchParams({
-        ...Object.fromEntries(request.body.entries()),
-        ...authParams,
-      });
+    if (this.authType === AuthenticationType.KEY_SECRET) {
+      request.type = ContentType.FORM_URLENCODED;
```
dragonmantank: Key/Secret does not automatically make our requests URL encoded

dragonmantank (changes requested):
Questions on access visibility and some code changes

packages/server-client/lib/enums/AuthenticationType.ts
manchuck: Since we opened up `authType` to be public, I want to highlight that this should not be used. Maybe we can change the verbiage? @pardel any suggestions?

packages/server-client/lib/fileClient.ts
manchuck: This change came about so we can use `ServerClient` standalone. Because this class extends `ServerClient`,  the visibility needs to stay the same. Making the `authType` public is the path of least resistance.

packages/server-client/lib/fileClient.ts
manchuck: you know I have not tested this out with that case. Lets file that as a seperate issue and address after this PR merges in

packages/server-client/lib/enums/AuthenticationType.ts
manchuck: We devided to leave this as is for now and will update when all doc blocks PRs have been closed

packages/server-client/lib/client.ts
manchuck: You can now set the body with GET and DELETE

packages/server-client/lib/fileClient.ts
manchuck: This is now async safe

packages/server-client/__tests__/__dataSets__/index.ts
```diff
@@ -7,26 +7,26 @@ import patchTests from './patch';
 import deleteTests from './delete';
 
 export const requestTests = [
-  {
-    label: 'Get requests',
-    tests: getTests,
-  },
+//  {
```
pardel: commented tests?! 
woooah

packages/server-client/lib/types/APILink.ts
```diff
@@ -1,4 +1,13 @@
+/**
+ * Represents a link object in the HAL format.
+ *
+ * @see {@link https://tools.ietf.org/html/rfc5988} for more details on Web Linking.
+ *
+ */
 export type APILink = {
```
pardel, line 7: Do we really need a type for this?

packages/server-client/package.json
```diff
@@ -11,7 +12,16 @@
     "url": "git+https://github.com/Vonage/vonage-node-sdk.git"
   },
   "license": "Apache-2.0",
-  "author": "Chris Tankersley <chris@ctankersley.com>",
+  "contributors": [
```
pardel, line 15: Please add github profiles rather than emails.

pardel (changes requested):

packages/server-client/lib/types/APILink.ts
manchuck: Yes this type is needed for when we are paging in some API's. For example, the meetings API needs to read in the link to determine if the last page has been reached:

https://github.com/Vonage/vonage-node-sdk/blob/3.x/packages/meetings/lib/meetings.ts#L82

packages/server-client/__tests__/__dataSets__/index.ts
manchuck: my bad did'nt push up after I uncommented

dragonmantank (approved):

packages/server-client/package.json
manchuck: Will address once all PRs for doc blocks have merged in

PR #884, opened 2023-11-01
docs(numbers): adding doc blocks

manchuck:
## Types of changes

## Checklist


maxkahan (approved):

packages/numbers/package.json
```diff
@@ -11,7 +12,20 @@
     "url": "git+https://github.com/Vonage/vonage-node-sdk.git"
   },
   "license": "Apache-2.0",
-  "author": "Kelly J Andrews <kelly@kellyjandrews.com>",
```
pardel, line 14: please change to github profiles instead of email addresses.

packages/numbers/lib/numbers.ts
```diff
@@ -104,64 +91,90 @@ const sortFeatures = (features: string[]): string => {
...
   public async buyNumber(
-    params?: NumbersParams,
+    params: NumbersParams,
   ): Promise<NumbersEmptyResponse> {
-    const mapping = { target_api_key: 'targetApiKey' };
-    const data = remapObjects(mapping, {}, params);
     const resp = await this.sendFormSubmitRequest<NumbersEmptyResponse>(
       `${this.config.restHost}/number/buy`,
-      data,
+      Client.transformers.snakeCaseObjectKeys(params) as Record<string, string>,
```
pardel, line 111: Can we not complicate things like this? Code should be written to be easily readable rather than look smart. What is the benefit of adding transformers here? Why not just a simple inline mapping?

pardel (changes requested):

packages/numbers/package.json
manchuck: Will resolve once all Doc Block PRs are merged in

packages/numbers/lib/numbers.ts
manchuck: This is done to normalize out responses and requests. Since our API's do not follow a consistent casing scheme (most use `snake_case`, some have `kebab-case`), we convert all responses to be `camelCase` for ease of use. 

In the cases where `kebab-case` is used, a developer would have to access the object like `resp.data['content-type]` rather than `resp.data.contentType`.

Camel Case was chosen since the popular linter rules require `camelCasing` for variable names (Google,  React, Airbnb) 

Using the transformer instead of defining the mapping, alleviates the need to push updates when new variables are introduced and prevent keys from being improperly mapped through fat fingering. If the API gets an update, the user will be able to access the new data without waiting for a new release of the SDK. We do map the response in `verifyv2` which is why there has to be work done to enable silent auth. 

PS for users of Typescript, we export all the types in our packages to allow for this use case. If there is an update to the API they can simple merge of the types with the new propert(ies) added.

PR #889, opened 2023-11-21
docs(sms): adding doc blocks

manchuck:
## Types of changes

## Checklist


packages/sms/lib/enums/SMSErrors.ts
```diff
@@ -1,22 +1,112 @@
...
+   */
+  INTERNAL_ERROR = '5',
+
+  /**
+   * Invalid Message: The platform was unable to process the message.
+   */
+  INVALID_MESSAGE = '6',
+
+  /**
+   * Barred Number: The number you're trying to send messages to is blacklisted.
```
superchilled:
```suggestion
   * Barred Number: The number you're trying to send messages to is blocklisted.
```

Would change this to blocklisted. I know we use 'blacklisted' in the actual docs, but we probably need to change that too.

packages/sms/lib/enums/SMSErrors.ts
```diff
@@ -1,22 +1,112 @@
...
+   * Invalid Network Code: Network code is unrecognized or doesn't match the destination.
+   */
+  INVALID_NETWORK_CODE = '22',
+
+  /**
+   * Invalid Callback URL: Callback URL is too long or contains illegal characters.
+   */
+  INVALID_CALLBACK_URL = '23',
+
+  /**
```
superchilled, line 98: Is it possible to change the actual enum value? Would suggest maybe `NON_ALLOWLISTED_DESTINATION`, and also change the doc comment accordingly.

packages/sms/lib/enums/SMSStatus.ts
```diff
@@ -1,22 +1,112 @@
+/**
+ * Enumeration representing possible SMS status codes.
```
superchilled, line 2: See review comments in `SMSErrors.ts`

packages/sms/lib/types/InboundMessage.ts
```diff
@@ -0,0 +1,86 @@
...
+   */
+  messageId: string;
+
+  /**
+   * The message body for this inbound message.
+   */
+  text: string;
+
+  /**
+   * The format of the message body.
```
superchilled, line 35: Would it be worth referencing the enum used for this?

packages/sms/lib/types/InboundMessage.ts
```diff
@@ -0,0 +1,86 @@
...
+   */
+  type: string;
+
+  /**
+   * The first word in the message body, converted to uppercase.
+   */
+  keyword: string;
+
+  /**
+   * The time when Vonage started to push this Delivery Receipt to your webhook endpoint.
```
superchilled:
```suggestion
   * The time when Vonage started to push this Inbound SMS to your webhook endpoint.
```

I think this is probably a typo/copy-and-paste-error in the docs

packages/sms/lib/types/Message.ts
```diff
@@ -0,0 +1,35 @@
+import { SMSMessageResponse } from './Responses';
+
+/**
+ * Interface representing an SMS message.
```
superchilled: Isn't this the interface for the API response when sending and SMS message  rather than the interface representing an SMS message?

packages/sms/lib/types/Requests/SMSRequestBody.ts
```diff
@@ -0,0 +1,88 @@
...
+ * Describes the structure of the request body used when sending an SMS message.
+ */
+export type SMSRequestBody = {
+  /**
+   * The recipient's phone number in E.164 format.
+   */
+  to: string;
+
+  /**
+   * The sender's name or number, specified in E.164 format.
```
superchilled:
```suggestion
   * The Alphanumeric senderID (if supported for the destination) or virtual number (specified in E.164 format) that the SMS is being sent from.
```

packages/sms/lib/types/Requests/SMSRequestBody.ts
```diff
@@ -0,0 +1,88 @@
...
+
+  /**
+   * Boolean indicating if a Delivery Receipt is requested (optional).
+   */
+  'status-report-req'?: boolean;
+
+  /**
+   * The Data Coding Scheme value of the message (optional).
+   */
+  'message-class'?: MessageClassEnum;
```
superchilled, line 72: Is this intentionally duplicated (e.g. one with the enum as a value and one with `number`)?

packages/sms/lib/types/Requests/SMSRequestBody.ts
```diff
@@ -0,0 +1,88 @@
...
+
+  /**
+   * Your own reference for the message (optional).
+   */
+  'client-ref'?: string;
```
superchilled, line 77: Duplicated

packages/sms/lib/types/Responses/DeliveryReceipt.ts
```diff
@@ -0,0 +1,76 @@
...
+
+  /**
+   * When the Delivery Receipt was received from the carrier in YYMMDDHHMM format (optional).
+   */
+  scts?: string;
+
+  /**
+   * The status of the request (optional).
+   */
+  errCode?: string;
```
superchilled, line 45: Should this reference the `SMSErrors` enum?

superchilled (commented):
These mostly make sense to me, but I've added a few comments.

As a general comment, although I'm not really familiar with the TSDoc syntax, I was wondering whether the `@remarks` tag would have been appropriate in a few more places?

packages/sms/lib/types/Responses/DeliveryReceipt.ts
manchuck: I chose to leave this as a string to allow for future proofing. If we get a new status that is in the request, then that could break TS compiling.

packages/sms/lib/enums/SMSErrors.ts
manchuck: I took this from the docs so long ago that it cannot be changed without making this a major release. In TS, you can't have two different keys have the same value in an ENUM. I can add a reference in the block.

packages/sms/lib/types/Message.ts
manchuck: Some context: In the past we never transformed responses. Now we convert all responses to `camelCase` for ease of use. Since this package was published with out the transformation, we keep the OG properties and add in the `camelCase`ed properties as well. This represents what we return back to the user after merging them together

superchilled (approved):

PR #907, opened 2024-01-28
fix: outdated/wrong algorithm argument passed to jsonwebtoken

Ymirke:
@vonage/jwt VerifySignature defaults to RS256. 

My best guess is that it's to accommodate a previous version of the authentication/authorization setup vonage had.

Either way, there is no reason for it, as JWTs contain the algorithm in the header.

Additionally, the JWTs I receive from Vonage are HS256. Meaning that this is breaking and verifyToken does not work.

There is another question as to wether this should be provided by the Vonage SDK, but that's up to you guys!

## Description

Removed RS256 argument passed to verify function from jsonwebtoken.

## Motivation and Context

My webhooks recieve HS256 tokens, not RS256 from Vonage. Either way it shouldn't be defaulted as the algorithm is contained in the JWT header.

## Testing Details

No test run done, should pass though.

--

--

## Types of changes

## Checklist

### DIFF
--- a/packages/jwt/lib/verifySignature.ts
+++ b/packages/jwt/lib/verifySignature.ts
@@ -44,9 +44,7 @@ export const verifySignature = (
   }
 
   try {
-    verify(jwt, privateKey, {
-      algorithms: ['RS256'],
-    });
+    verify(jwt, privateKey);
     return true;
   } catch (error) {
     log('Error when verifying token', error);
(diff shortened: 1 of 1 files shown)


manchuck:
@Ymirke sorry about that. Do you mind updating the algorithm instead of removing it? We want to be explicit in what we accept

manchuck (approved):

PR #956, opened 2024-09-05
fix applications response types

s-lukashenka-micoworks:
- `Applications` API response type is missing a `privateKey` field.
- `Applications` API does not require all the three webhook URLs to be passed during creation but the type does not allow to omit any.

## Motivation and Context

This PR helps resolve failing type checks. The API itself works fine and has the mentioned field in its response.
Related to https://github.com/Vonage/vonage-node-sdk/issues/955.

## Testing Details

Not sure if I did things right but I ran the test suite on the PR branch.
Please, let me know if some other tests or changes are needed.

## Types of changes

## Checklist

### DIFF
--- a/packages/applications/lib/types/Application.ts
+++ b/packages/applications/lib/types/Application.ts
@@ -36,7 +36,7 @@ export type Application = {
      */
     publicKey?: string;
     /**
-     * The public key for the application.
+     * The private key for the application.
      *
      * @remarks
      * Store this value securely.
--- a/packages/applications/lib/types/Application.ts
+++ b/packages/applications/lib/types/Application.ts
@@ -35,6 +35,14 @@ export type Application = {
      *
      */
     publicKey?: string;
+    /**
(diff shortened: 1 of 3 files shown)


packages/applications/lib/types/Application.ts
```diff
@@ -35,6 +35,14 @@ export type Application = {
      *
      */
     publicKey?: string;
+    /**
+     * The private key for the application.
+     *
+     * @remarks
+     * Store this value securely.
```
manchuck: Update to be: 

`This will only be present when the application is created, or when you cycle the public key when doing an update`

manchuck (changes requested):

packages/applications/lib/types/Application.ts
s-lukashenka-micoworks: @manchuck updated the doc 🙇

manchuck (approved):

PR #993, opened 2025-05-12
feat: add more language codes for Captions API 

behei-vonage:
This PR adds more language codes for Captions API as it currently lists only `en-us`.
The full list is here: https://tokbox.com/developer/rest/#starting-live-captions

## Description
Not too many details can be added, just listing out more languages that are supported in the Captions API. 

## Motivation and Context
This PR adds on additional languages to the captions API. 

## Testing Details
Additional languages are supported when trying the captions API. 

## Types of changes

- [👍 ] New feature (non-breaking change which adds functionality)

## Checklist
- [X] My code follows the code style of this project.
- [X] I have read the **CONTRIBUTING** document.
- [n/a] I have added tests to cover my changes.
- [n/a] All new and existing tests passed.

### DIFF
--- a/packages/video/lib/types/CaptionOptions.ts
+++ b/packages/video/lib/types/CaptionOptions.ts
@@ -5,7 +5,7 @@ export type CaptionOptions = {
   /**
    * The language code for captions (e.g., "en-us").
    */
-  languageCode?: 'en-us';
+  languageCode?: 'en-US' | 'en-AU' | 'en-GB' | 'zh-CN' | 'fr-FR' | 'fr-CA' | 'de-DE' | 'hi-IN' | 'it-IT' | 'ja-JP' | 'ko-KR' | 'pt-BR' | 'th-TH';
 
   /**
    * The maximum duration for captions.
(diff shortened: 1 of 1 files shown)


packages/video/lib/types/CaptionOptions.ts
```diff
@@ -5,7 +5,7 @@ export type CaptionOptions = {
   /**
    * The language code for captions (e.g., "en-us").
    */
-  languageCode?: 'en-us';
+  languageCode?: 'en-US' | 'en-AU' | 'en-GB' | 'zh-CN' | 'fr-FR' | 'fr-CA' | 'de-DE' | 'hi-IN' | 'it-IT' | 'ja-JP' | 'ko-KR' | 'pt-BR' | 'th-TH';
```
behei-vonage: also note that passing in `en-us` does not currently work as `US` has to be uppercase.
for example, I'm not able to `enableCaptions` if I pass in the `languageCode` set to `en-us`; however, I have no issue if I pass it in without one

manchuck (approved):

packages/video/lib/types/CaptionOptions.ts
manchuck: Good catch.

PR #1039, opened 2026-01-14
refactor: maintainance fixes

manchuck:
Updated all packages
Removed lodash dependenancies
Removed support for sending API Key/Secret through the query string and message body


alnacle (approved):
LGTM

PR #1295, opened 2026-07-21
feat: added unsupported network type for identity status

manchuck:
## Description

This is a quick addition to the Identity Insights Status, along with fixing the type returned for Typescript users

## Types of changes

## Checklist

### DIFF
--- a/packages/identity-insights/lib/enums/codeStatus.ts
+++ b/packages/identity-insights/lib/enums/codeStatus.ts
@@ -5,54 +5,59 @@ export enum Status {
   /**
    * The country or mobile network is not supported by available suppliers.
    */
-  NO_COVERAGE = 'no_coverage',
+  NO_COVERAGE = 'NO_COVERAGE',
 
   /**
    * The purpose used is not valid or allowed for this Insight.
    */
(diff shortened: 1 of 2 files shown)


packages/identity-insights/lib/enums/codeStatus.ts
```diff
@@ -5,54 +5,59 @@ export enum Status {
   /**
    * The country or mobile network is not supported by available suppliers.
    */
-  NO_COVERAGE = 'no_coverage',
+  NO_COVERAGE = 'NO_COVERAGE',
```
dragonmantank, line 8: Ugh, I hate to ask, but I should:

This is a breaking change. Did we intend that, or are we correcting a bug? Or do we need to do this in a BC way?

dragonmantank (changes requested):

packages/identity-insights/lib/enums/codeStatus.ts
manchuck: This is correcting a bug. The API spec defines this value in ALL_CAPS

dragonmantank (approved):

